PR #1, opened 2022-08-16
Double trimming

theis999:
Adds trimming of tracing zeroes when extracting x-cord, y-cord & units. from an existing task to the task designer.


MortenTobiasNielsen:
@theis999 Thanks for your contributions. I am having a little trouble upgrading wxWidgets to the latest version so e.g. ToInt is available, so it might first be tomorrow that I get this merged in.

PR #3, opened 2022-08-16
Resume from save

theis999:
Adds an option to resume from a save point without interfering with the building list.


MortenTobiasNielsen:
I like the thought behind this, but I think I will call it Start instead - as in start the steps here. :) Thanks for the help. I hope you will come with more valuable feedback.

PR #9, opened 2022-08-18
Coordinate comparison

theis999:
Makes the match building from task list, to building list slightly more robust.

### DIFF
--- a/cMain.cpp
+++ b/cMain.cpp
@@ -988,7 +988,7 @@ void cMain::OnChangeTaskClicked(wxCommandEvent& event) {
 
 		for (int i = 0; i < std::stoi(building_amount_of_buildings); i++) {
 			for (int j = row_num + 1; j < total_rows; j++) {
-				if (grid_tasks->GetCellValue(j, 1) == building_x_cord && grid_tasks->GetCellValue(j, 2) == building_y_cord) {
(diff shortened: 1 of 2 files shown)


theis999:
It fixes a few comparisons, which got damaged due to the trimming of tracing zeroes. It incidentally also fixes the odd case of a building be placed in ( -0.0 ) and trying to interact with it in ( 0.0 ).
attaching a test fil:
[test.txt](https://github.com/MortenTobiasNielsen/TAS-Helper-for-Factorio/files/9373994/test.txt)
e

MortenTobiasNielsen:
I have made another implementation of the trimming, so solve the problem - didn't know that you would make a fix for it. Could you pull from main and see if my fix solve the problem without all this extra code?

I can see the situation where you might get a strange error due to this, but it is a very niche case where the user can double click the building to get the coordinates.

theis999:
Your implementation solved some of the issues. However not all of them. In particular when interacting with the last built building, you get warnings all the time. The appears to come from when comparing the trimmed value of the field with the raw value.

MortenTobiasNielsen:
Do you have a specific example of when it gives errors, I cannot reproduce it with a rudimentary example like building a stone furnace at 10.0, 5.0 and then putting wood into it.

theis999:
[test.txt](https://github.com/MortenTobiasNielsen/TAS-Helper-for-Factorio/files/9374759/test.txt)
In this scenario, if i move task 6 up so it disconnects with task 5 and then move it down again. It should reconnect with task 5 but doesn't.

Issue #24, opened 2022-08-31
Goal: Debug


theis999:
It would be nice to have debug / development goal, like any% or steelaxe%. Where additional information is printed in the console for easier understanding of what is going on.

MortenTobiasNielsen:
What level of detail would you like to have? E.g. printing every tick seems a bit excessive.

theis999:
I am not entirely sure yet. I know i want to know which tick an action begins at so i can calculate which tick a smelt or similar is completed.

MortenTobiasNielsen:
That sounds like a good idea.

MortenTobiasNielsen:
This would take a lot of work, but what if you were able to give a command from the tool, which would set a timer for when the next unit is done from a furnace, assembling machine, etc. which would then be shown as an ordered list in game as a GUI element. The timer would need to be able to cycle by itself, so the user doesn't need to constantly set timers. 

This would also be very useful for crafts.

theis999:
It might be easier to write a supporting mod, that prints information on entities relating to their status. Similar to what [Bottleneck](https://mods.factorio.com/mod/Bottleneck)

MortenTobiasNielsen:
I think you are right.

PR #28, opened 2022-09-03
Debug goal

theis999:
#24 Goal debug added.

Added Goal->Debug in the GUI
Added functions msg & warning to control_lua1, to display messages in slightly different ways
Changed some debug messages to either use msg or warning.
Changed that debug messages would be show at goals other than debug.

Fixed mining rock event - including that Big sandy rock would yield (was 20stone) a below average amount (now 24stone).

### DIFF
--- a/control_info.h
+++ b/control_info.h
@@ -48,9 +48,29 @@ local neg_neg = false
 local drop_item
 local drop_position
 
+--Print message intended for viewers
+local function msg(msg) 
+    player.print(msg)
+end
+
+--Print debug message about what the tas is doing
 local function debug(msg)
 	if debug_state then
 		player.print(msg)
+        player.print(string.format(
+            "Seconds: %s, tick: %s, player position [%d, %d]",
+            game.tick / 60,
+            game.tick,
+            player.position.x,
+            player.position.y
+	))
+	end
+end
+
+--Print warning in case of errors in tas programming
+local function warning(msg)
+    if debug_state then
+		player.print(msg, {r=1, g=1}) -- print warnings in yellow
 	end
 end
 
--- a/control_info.h
+++ b/control_info.h
@@ -775,26 +775,38 @@ script.on_event(defines.events.on_tick, function(event)
 	end	
 end)
 
+local function mining_event_replace(event, item_name, amount)
+	local count = event.buffer.get_item_count(item_name)
+	if count < amount then
+		event.buffer.insert({name=item_name, count=amount-count})
+	elseif count > amount then
+		event.buffer.remove({name=item_name, count=count-amount})
+	end --on correct amount do nothing
+end
+
 script.on_event(defines.events.on_player_mined_entity, function(event)
 
 	if (steps[step][1] == "break" or steps[step][2] == "stop") then
 		return
 	end
-
+	if event.entity.name == "rock-huge" or event.entity.name == "rock-big" or event.entity.name == "sand-rock-big" then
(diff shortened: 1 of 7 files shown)


control_info.h
```diff
@@ -47,20 +47,40 @@ local neg_neg = false
...
+            "Seconds: %s, tick: %s, player position [%d, %d]",
+            game.tick / 60,
+            game.tick,
+            player.position.x,
+            player.position.y
+	))
+	end
+end
+
+--Print warning incase of errors in tas programming
```
MortenTobiasNielsen, line 69: in case

control_info.h
```diff
@@ -47,20 +47,40 @@ local neg_neg = false
...
 local function save(nameOfSaveGame)
 	if game.is_multiplayer() then
-		debug(string.format("Task: %s, saving game as %s", task[1], nameOfSaveGame))
+		msg(string.format("Task: %s, saving game as %s", task[1], nameOfSaveGame))
 		game.server_save(nameOfSaveGame)
 		return true
 	end
 
-	debug(string.format("Task: %s, saving game as _autosave-%s", task[1], nameOfSaveGame))
+	warning(string.format("Task: %s, saving game as _autosave-%s", task[1], nameOfSaveGame))
```
MortenTobiasNielsen, line 83: Shouldn't this also be a msg instead of a warning?

control_info.h
```diff
@@ -704,23 +726,22 @@ script.on_event(defines.events.on_tick, function(event)
...
+						warning(string.format("Task: %s, Action: %s, Step: %s - Mine: Cannot reach resource", steps[step][1][1], steps[step][1][2], step))
 						debug_state = false
 					else
 						mining = 0
 					end
 				end
 				
 			elseif doStep(steps[step]) then
-				-- Do step while standing still
+                --debug(string.format("Task: %s, Action: %s, Step: %s - do step while standing still", steps[step][1][1], steps[step][1][2], step))
```
MortenTobiasNielsen, line 737: Remove this line.

control_info.h
```diff
@@ -704,23 +726,22 @@ script.on_event(defines.events.on_tick, function(event)
...
 				step = step + 1
 
 			end
 		else
 			if steps[step][2] ~= "walk" and steps[step][2] ~= "mine" and steps[step][2] ~= "idle" and steps[step][2] ~= "pick" then
 				if doStep(steps[step]) then
-					
-					-- Do step while walking
+					--debug(string.format("Task: %s, Action: %s, Step: %s - do step while walking", steps[step][1][1], steps[step][1][2], step))
```
MortenTobiasNielsen, line 744: Remove this line.

MortenTobiasNielsen (commented):
Good work, only have a few small comments.

control_info.h
theis999: I had the intention to log if a step was done while walking. Initial testing showed that was far far to annoying so i commented it out. I shouldn't have included it in the commit.

control_info.h
theis999: You are proably correct, it effectively doesn't matter since there shouldn't be any saves during the final run. I will however change it.

MortenTobiasNielsen: It was mostly to be aligned with msg being used when it is a multi-player game.  

I would guess that removing the _Save_ tasks would offset the timings of the game ever so slightly, which could be a problem, but I have not tested it.

MortenTobiasNielsen (approved):
Nice - thanks for the contribution. :)

PR #32, opened 2022-09-04
display floating txt when adding or removing items

theis999:


control_info.h
```diff
@@ -134,7 +134,11 @@ local function put()
 		debug(string.format("Task: %s, Action: %s, Step: %d - Put: Entity is already full", task[1], task[2], step))
 		return true
 	end
-
+	if item ~= nil then
+		-- -2 iron plate (5)
+		local t = string.format("+%d %s (%d)", amount, item, amount + player.get_item_count(item))
```
MortenTobiasNielsen, line 139: This doesn't work as I would expect. When I put 1 iron ore into a furnace it writes "+1 iron-ore (2)" even though it is the first iron-ore given to that furnace and when I put in 6 automation-science-packs it wrote 16. I guess it is because the 10 I had in my inventory before I put the 6 are added together instead of getting what was in the entity before and then adding that to what is put into the entity.

MortenTobiasNielsen: I would also like the amount of floating txt to be aligned with what is shown in game. I don't seem to be able to recreate that floating txt is being shown at the players location to indicate how much is in the players inventory after putting/taking entities.

theis999: It might not be intuitive, but i recreated what i saw in the base game. Where it display (item change, item name, item count in inventory) at the location of the target inventory.
I am open for suggestions that deviate from the regular factorio experience, if they make the tas experience better.

I took some images from the base game to see show what i designed it to do:

image
image

theis999: The text managed to float quite far before i could grab the images but it started at my cursor. In the first image you can also see that i had 5 wood left in inventory after the transfer.

MortenTobiasNielsen: I though for a moment that it showed the amount of units in the entity and not what is left in the players inventory, but it makes sense that it is the other way around. Then I think this:
```
local t = string.format("+%d %s (%d)", amount, item, amount + player.get_item_count(item))
```
needs to be changed to this:
```
local t = string.format("+%d %s (%d)", amount, item, amount - player.get_item_count(item))
```
but I think it is cleaner to move this below player.remove_item{name=item, count=amount}, so player.get_item_count has the correct amount without having to calculate. 

The same should be done for taking.

control_info.h
```diff
@@ -163,6 +167,11 @@ local function take()
 		return true
 	end
 
+	if item ~= nil then
+		-- +2 iron plate (5)
+		local t = string.format("+%d %s (%d)", amount, item, amount + player.get_item_count(item))
+		player.create_local_flying_text{text=t, position=player_position, time_to_live=120}
```
MortenTobiasNielsen, line 173: When items are taken from entities the floating txt is located at the entity where the items are taken from, not at the player. 

You might be able to set create_at_cursor to true (and ignore position and time_to_live) depending on whether the cursor is actually moved around when executing tasks. 
image

control_info.h
MortenTobiasNielsen: I though for a moment that it showed the amount of units in the entity and not what is left in the players inventory, but it makes sense that it is the other way around. Then I think this:

local t = string.format("+%d %s (%d)", amount, item, amount + player.get_item_count(item))
needs to be changed to this:

local t = string.format("+%d %s (%d)", amount, item, amount - player.get_item_count(item))
but I think it is cleaner to move this below player.remove_item{name=item, count=amount}, so player.get_item_count has the correct amount without having to calculate.

The same should be done for taking.

MortenTobiasNielsen: I created a comment, but by mistake deleted it. I hope you have it in your inbox. :)

theis999: I see it yes. I haven't yet decided if i agree, but i think we also need to [can_insert](https://lua-api.factorio.com/latest/LuaControl.html#LuaControl.can_insert) to make sure the items can actually be added.

control_info.h
theis999: It is appears that you caught a mistake of copy pasting code. The position was supposed to target_position just like insert.

I remember trying create_at_cursor and it not behaving how i wanted it too. Looking at the api description again, it might have been because the text i wanted to show was to short. I will give it a second try.

theis999: The create_at_cursor spawns the text at the cursor, obviously 😎. But the TAS doesn't move the cursor, so the effect is very confusing.
I removed the time_to_live and that seems to be an improvement.

control_info.h
```diff
@@ -187,6 +190,9 @@ local function take()
 
 	player.play_sound{path="utility/inventory_move"}
```
MortenTobiasNielsen, line 191: We might want to move this and the floating txt down below the check of amount - I think it will be confusing to hear inventory movement and see floating txt and then get a warning that it wasn't possible to take anything.

theis999: Correct. I moved the play sound and display inventory transfer to the end of the functions.

MortenTobiasNielsen (approved):

PR #37, opened 2022-09-05
Moved buildings grid to the left pane

### LINKED ISSUE #23: Move buildings tab to the left pane
I think it would make more sense from users point of view, to have buildings tab in the left pane with groups and templates.

MortenTobiasNielsen:
closes #23


MortenTobiasNielsen:
@theis999 Would you review this PR? :)

GUI_Base.cpp
```diff
@@ -817,7 +894,7 @@ GUI_Base::GUI_Base(wxWindow* parent, wxWindowID id, const wxString& title, const
 	m_panel3->SetSizer(bSizer561);
 	m_panel3->Layout();
 	bSizer561->Fit(m_panel3);
-	m_notebook1->AddPage(m_panel3, wxT("Group"), true);
+	m_notebook1->AddPage(m_panel3, wxT("Group"), false);
```
theis999, line 897: I don't understand this change

theis999 (approved):
Only a minor comment

GUI_Base.cpp
MortenTobiasNielsen: I don't really understand why it is needed either, but it is part of the generated code from WxFromBuilder.

PR #38, opened 2022-09-05
Changed control_info to add sounds when placing tiles.

### LINKED ISSUE #25: Placement sounds
It appears that entities place through the API doesn't generate the usual game sounds.
useful references: 

- https://lua-api.factorio.com/latest/LuaPlayer.html#LuaPlayer.play_sound
- https://lua-api.factorio.com/latest/events.html#on_built_entity

also 
_Oh yeah I also thought of that when I was watching some time ago. Also good would be having graphics show when putting/taking items, which should be possible with https://lua-api.factorio.com/latest/LuaPlayer.html#LuaPlayer.create_local_flying_text_

MortenTobiasNielsen:
closes #25 

Sets run = true as the default, so the steps executes until release is written in the console.


MortenTobiasNielsen:
@theis999 please review this PR

control_info.h
```diff
@@ -12,7 +12,7 @@ std::string info = { R"info(
 std::string control_lua1 = R"control_lua1(
 local steps = require("steps")
 local debug_state = true
-local run = false
+local run = true
```
theis999, line 15: 👍

control_info.h
```diff
@@ -250,8 +250,18 @@ local function build()
 		if item_is_tile(item) then
 			if item == "stone-brick" then 
 				player.surface.set_tiles({{position = target_position, name = "stone-path"}})
-            else 
+				player.surface.play_sound{path="tile-build-small/stone-path", position=target_position}
+			-- something can be done here with item:gsub, but I couldn't get it to work - kebt getting refined-d-concrete when using item:gsub("hazard-", "")
```
theis999, line 254: It looks to me like [there is effectively 2 place tile sounds](https://github.com/wube/factorio-data/blob/master/base/prototypes/tile/tiles.lua). Landfill and "concrete", where concrete is all the concrete types and stone path. Which also lines up with sound files are actually played, which you can see through _f4->show-recently-played-sound-info_.

theis999 (changes requested):
You over complicated it a bit

control_info.h
MortenTobiasNielsen: It also felt overly complicated. :) So I should pretty much play stone-path unless the item is landfill?

theis999: That appears to be what the base game does, so yes.

theis999 (approved):
Looks perfect

MortenTobiasNielsen:
It unfortunately doesn't work as intended. There is no check whether the players should actually be able to place tiles at the position specified.

theis999:
I see, are you working on it right now? or would you rather have me take a look at it.

MortenTobiasNielsen:
It would be nice if you would take a look at it - I don't have more time today.

MortenTobiasNielsen:
@theis999 I think I have fixed it now and it also works with hazard variation.

control_info.h
```diff
@@ -244,8 +244,8 @@ end
 local function tile_is_in_reach()	
 	local x = player.position.x - target_position[1]
 	local y = player.position.y - target_position[2]
-	local dis = math.sqrt(x^2+y^2) --sqrt(a^2+b)^=sqrt(c^2)
-	return dis < 11 -- reach is 10 tiles
+	local dis = math.sqrt(x^2+y^2) --sqrt(a^2+b^2)=sqrt(c^2)
+	return dis <= 10.25 -- It seems like 10.25 aligns best with the current walking algorithm
```
theis999, line 248: Good use of comment to explain a magic number

control_info.h
```diff
@@ -264,24 +264,32 @@ local function build()
 
 	if (item ~= "rail") then
 		if item_is_tile(item) then
-			if not tile_is_in_reach() then 
+			if tile_is_in_reach() then
+				if item == "stone-brick" then 
+					player.surface.set_tiles({{position = target_position, name = "stone-path"}})
+				elseif (item == "hazard-concrete") then
+					player.surface.set_tiles({{position = target_position, name = item.."-left"}})
```
theis999, line 271: It would be more correct to use building direction to set hazard direction.

control_info.h
```diff
@@ -264,24 +264,32 @@ local function build()
...
+				elseif (item == "refined-hazard-concrete") then
```
theis999, line 272: It is possible to combine refined-hazard and hazard.

control_info.h
theis999: but on the other hand, i don't think hazard concrete will ever be used in a tas

theis999 (approved):
Good job. If you merge two hazard types then i approve.

control_info.h
MortenTobiasNielsen: I did a little testing, but couldn't really see that hazard had a direction. It seems like it turns the same way and it would be difficult to use building direction, given that it has four (north, south, east, and west), but hazard only have left and right. 

I think you are right, in a TAS it (or any speed run for that matter) it would be a waste of time to use hazard.

control_info.h
MortenTobiasNielsen: Good addition - I have changed it.

theis999 (approved):

PR #39, opened 2022-09-06
Fixed a few issues

MortenTobiasNielsen:


MortenTobiasNielsen:
@theis999

EZRaiderz-TAS-Helper.vcxproj
```diff
@@ -42,7 +42,7 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v142</PlatformToolset>
+    <PlatformToolset>v143</PlatformToolset>
```
theis999, line 45: I see you upgraded to vs2022

cMain.cpp
```diff
@@ -3224,7 +3224,7 @@ std::string cMain::extract_y_cord() {
 std::string cMain::extract_units() {
 	float units = wxAtof(txt_units->GetValue());
 	
-	if (units < 0 && (rbtn_rotate->GetValue() || rbtn_idle->GetValue())) {
+	if (units < 1 && (rbtn_rotate->GetValue() || rbtn_idle->GetValue())) {
```
theis999, line 3227: It is always interesting when magic-values change. But if you think it is an improvement, then it is fine.

control_info.h
```diff
@@ -546,7 +546,9 @@ local function launch()
 
 	return player_selection.launch_rocket()
 end
+)control_lua1";
 
+std::string control_lua2 = R"control_lua2(
```
theis999, line 551: 👌

theis999 (changes requested):
I typically like to separate different issues into different commits. This has the dual purpose of having more concise commit messages and more specific git logs per file/line.
Here i have to read the git branch to even understand what is going on.

I would rather see you separate this into 2 commits:
1. To fix build issues
2. To fix rotate bug

The actual changes seem fine.

1.

cMain.cpp
MortenTobiasNielsen: The application would try to rotate buildings "All" amount of times, if the units wasn't specified, which would break the application, I do though agree that magic variables isn't a good thing to have. :)

EZRaiderz-TAS-Helper.vcxproj
MortenTobiasNielsen: Yeah I thought it would make sense to do - vs2022 are relatively mature now.

PR #40, opened 2022-09-06
Added geometric function to determine distance

theis999:


control_info.h
```diff
@@ -241,6 +241,13 @@ local function item_is_tile(item)
     return false
 end
 
+local function tile_is_in_reach()	
+	local x = player.position.x - target_position[1]
+	local y = player.position.y - target_position[2]
+	local dis = math.sqrt(x^2+y^2) --sqrt(a^2+b)^=sqrt(c^2)
```
MortenTobiasNielsen, line 247: --sqrt(a^2+b^2)=sqrt(c^2)

control_info.h
```diff
@@ -241,6 +241,13 @@ local function item_is_tile(item)
...
+	return dis < 11 -- reach is 10 tiles
```
MortenTobiasNielsen, line 248: I think Return dis <= 10 would be more accurate

control_info.h
```diff
@@ -257,13 +264,17 @@ local function build()
 
 	if (item ~= "rail") then
 		if item_is_tile(item) then
+			if not tile_is_in_reach() then
```
MortenTobiasNielsen, line 267: We need to check if the player is currently walking and first give this warning, when that is not the case. Like it is done for placing buildings.

MortenTobiasNielsen (changes requested):

control_info.h
theis999: Yeah, my bad.

control_info.h
theis999: Fair, i don't have a preference either way.

control_info.h
theis999: You are right

control_info.h
MortenTobiasNielsen: It is not so much about preference as it is about what is correct game mechanics - I want it to be true to the game. 😊

PR #43, opened 2022-09-08
Set default value for take&put task

theis999:
Adds default value to to/from field when choosing either take or put task. Which should make it slightly easier to use.

Request from zaspar:
_do you think it's possible to have the from/into box to default to output when using 'take' and input when using 'put'?
atm it just defaults to whatever I last used
so if I do a take output then the next put will default to output which I will never need to use_


MortenTobiasNielsen (approved):
Nice change - will make it easier to use.

PR #45, opened 2022-09-09
Added auto-put functionality

MortenTobiasNielsen:
Made some restructuring to remove hard-coded strings

### DIFF
--- a/cMain.cpp
+++ b/cMain.cpp
@@ -916,7 +916,33 @@ void cMain::OnAddTaskClicked(wxCommandEvent& event) {
 		if (check_buildings_grid()) {
 			if (setup_for_task_group_template_grid()) {
 
-				update_tasks_grid();	
+				update_tasks_grid();
+
+				std::string base_units = "1";
+				std::string to_check = item;
+				string_capitalized(to_check);
+
+				if (task == struct_tasks_list.build) {
+					if (check_furnace->IsChecked() && to_check == struct_auto_put_furnace_list.stone || to_check == struct_auto_put_furnace_list.steel) {
+						auto_put(struct_fuel_list.coal, base_units, struct_from_into_list.fuel);
+					}
+
+					if (check_burner->IsChecked() && to_check == struct_auto_put_burner_list.burner_mining_drill || to_check == struct_auto_put_burner_list.burner_inserter || to_check == struct_auto_put_burner_list.boiler) {
+						auto_put(struct_fuel_list.coal, base_units, struct_from_into_list.fuel);
+					}
+
+					if (check_lab->IsChecked() && to_check == struct_science_list.lab) {
+						auto_put(struct_science_list.red_science, base_units, struct_from_into_list.input);
+					}
+				}
+
+				if (task == struct_tasks_list.recipe && check_recipe->IsChecked() ) {
+					std::vector<std::string> recipe = testing.find(to_check)->second;
+
+					for (int i = 0; i < recipe.size(); i += 2 ) {
+						auto_put(recipe[i], recipe[i + 1], struct_from_into_list.input);
+					}
+				}
 			}
 		} else {
 			if (row_num != grid_tasks->GetNumberRows()) {
@@ -2833,6 +2859,10 @@ std::string FormatString(wxString s) {
 		return s.ToStdString();
 	}
 
+	if (i == 0) {
+		return s.ToStdString() + ".0";
+	}
+
 	for (i; i > 0; i--) {
 		if (s[i] == '.') {
 			i += 2;
@@ -2857,14 +2887,14 @@ void cMain::update_parameters(wxGrid* grid, wxCommandEvent& event) {
 	std::string y_cord_formatted = FormatString(y_cord);
 	std::string units_formatted = FormatString(units);
 
-	if (task == "Game Speed") {
+	if (task == struct_tasks_list.game_speed) {
 		OnGameSpeedMenuSelected(event);
 		txt_units->SetValue(units_formatted);
 		
 		return;
 	}
 	
-	if (task == "Walk") {
+	if (task == struct_tasks_list.walk) {
 		OnWalkMenuSelected(event);
 		txt_x_cord->SetValue(x_cord_formatted);
 		txt_y_cord->SetValue(y_cord_formatted);
@@ -2873,7 +2903,7 @@ void cMain::update_parameters(wxGrid* grid, wxCommandEvent& event) {
 		return;
 	}
 	
-	if (task == "Mine") {
+	if (task == struct_tasks_list.mine) {
 		OnMineMenuSelected(event);
 		txt_x_cord->SetValue(x_cord_formatted);
 		txt_y_cord->SetValue(y_cord_formatted);
@@ -2882,7 +2912,7 @@ void cMain::update_parameters(wxGrid* grid, wxCommandEvent& event) {
 
 		return;
 	}
-	if (task == "Rotate") {
+	if (task == struct_tasks_list.rotate) {
 		OnRotateMenuSelected(event);
 		txt_x_cord->SetValue(x_cord_formatted);
 		txt_y_cord->SetValue(y_cord_formatted);
@@ -2892,7 +2922,7 @@ void cMain::update_parameters(wxGrid* grid, wxCommandEvent& event) {
 		return;
 	}
 	
-	if (task == "Craft") {
+	if (task == struct_tasks_list.craft) {
 		OnCraftMenuSelected(event);
 		txt_units->SetValue(units_formatted);
 		cmb_item->SetValue(item);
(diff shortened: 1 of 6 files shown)


cMain.cpp
```diff
@@ -916,7 +916,33 @@ void cMain::OnAddTaskClicked(wxCommandEvent& event) {
 		if (check_buildings_grid()) {
 			if (setup_for_task_group_template_grid()) {
 
-				update_tasks_grid();	
+				update_tasks_grid();
```
theis999, line 919: You are removing the same line your adding?

cMain.cpp
```diff
@@ -916,7 +916,33 @@ void cMain::OnAddTaskClicked(wxCommandEvent& event) {
...
+
+				std::string base_units = "1";
+				std::string to_check = item;
+				string_capitalized(to_check);
+
+				if (task == struct_tasks_list.build) {
+					if (check_furnace->IsChecked() && to_check == struct_auto_put_furnace_list.stone || to_check == struct_auto_put_furnace_list.steel) {
```
theis999, line 926: Why specifically stone and steel? and not copper and iron

cMain.cpp
```diff
@@ -916,7 +916,33 @@ void cMain::OnAddTaskClicked(wxCommandEvent& event) {
...
+						auto_put(struct_fuel_list.coal, base_units, struct_from_into_list.fuel);
+					}
+
+					if (check_burner->IsChecked() && to_check == struct_auto_put_burner_list.burner_mining_drill || to_check == struct_auto_put_burner_list.burner_inserter || to_check == struct_auto_put_burner_list.boiler) {
```
theis999, line 930: If i remember correctly, && is done before ||. So you may need a parentese to achieve the logic you want.

cMain.cpp
```diff
@@ -916,7 +916,33 @@ void cMain::OnAddTaskClicked(wxCommandEvent& event) {
...
+					if (check_lab->IsChecked() && to_check == struct_science_list.lab) {
+						auto_put(struct_science_list.red_science, base_units, struct_from_into_list.input);
+					}
+				}
+
+				if (task == struct_tasks_list.recipe && check_recipe->IsChecked() ) {
+					std::vector<std::string> recipe = testing.find(to_check)->second;
+
+					for (int i = 0; i < recipe.size(); i += 2 ) {
+						auto_put(recipe[i], recipe[i + 1], struct_from_into_list.input);
```
theis999, line 943: This took me a few minutes to comprehend 😎

cMain.cpp
```diff
@@ -2966,29 +2996,29 @@ void cMain::update_parameters(wxGrid* grid, wxCommandEvent& event) {
 		return;
 	}
 	
-	if (task == "Start") {
+	if (task == struct_tasks_list.start) {
```
theis999, line 3090: With this many if statements on the same variable, we should consider a switch instead.

utils.h
```diff
@@ -29,6 +29,64 @@ static const std::vector<std::string> splitter_list = { "Splitter",
...
+																		 { "Productivity module 2", {"Advanced circuit", "5", "Processing unit", "5", "Productivity module", "4" }},
+																		 { "Productivity module 3", {"Advanced circuit", "5", "Processing unit", "5", "Productivity module 2", "5" }},
+																		 { "Speed module", {"Advanced circuit", "5", "Electronic circuit", "5" }},
+																		 { "Speed module 2", {"Advanced circuit", "5", "Processing unit", "5", "Speed module", "4" }},
+																		 { "Speed module 3", {"Advanced circuit", "5", "Processing unit", "5", "Speed module 2", "4" }},
+																		 { "Automation science pack", {"Copper plate", "1", "Iron gear wheel", "1" }},
+																		 { "Logistic science pack", {"Inserter", "1", "Transport belt", "1" }},
+																		 { "Chemical science pack", {"Advanced circuit", "3", "Engine unit", "2", "Sulfur", "1" }},
+																		 { "Production science pack", {"Electric furnace", "1", "Productivity module", "1", "Rail", "30" }},
+																		 { "Utility science pack", {"Flying robot frame", "1", "Low density structure", "3", "Processing unit", "2" }},
```
theis999, line 86: How do you have the patience to type all of this in?

utils.h
```diff
@@ -675,9 +732,70 @@ static const std::vector<std::string> tech_list = { "Advanced electronics",
...
+struct {
+	std::string lab = "Lab";
+	std::string red_science = "automation-science-pack";
+	std::string green_science = "logistic-science-pack";
+	std::string black_science = "military-science-pack";
+	std::string blue_science = "chemical-science-pack";
+	std::string purple_science = "production-science-pack";
+	std::string yellow_science = "utility-science-pack";
+	std::string white_science = "space-science-pack";
+} struct_science_list;
```
theis999, line 798: Using structs to list constants seems wrong to me. I couldn't find any other way on google, so i guess it is how it is done. I want the enums from c# 😒

theis999 (changes requested):
I believe the logic in line 926 is wrong.

I have pointed out a few things that could maybe be done better, but since they were just updated and not made this way in this commit, it is fine to leave it for now.

theis999 (approved):

theis999 (approved):

theis999 (changes requested):
It feels like i have read this code before  😒 
Anyway using an if statement to set a bool is obviously suspect

cMain.cpp
MortenTobiasNielsen: This is to check if the building being build is a stone/steel furnace and then creating a put coal task if that is the case. It is not to put stone/steel/copper/iron into a furnace or set a recipe for it. :)

cMain.cpp
MortenTobiasNielsen: No clue why github thinks I changed anything here.

cMain.cpp
MortenTobiasNielsen: I have tested the logic with all three and it works. It should go "is it a burner?" and then check each of the "or" statements afterwards.

cMain.cpp
MortenTobiasNielsen: Understandable it took me a good amount more than a minute to figure out. :D 

It might have helped a little if I had renamed my "testing" object. :)

cMain.cpp
MortenTobiasNielsen: I am not too familiar with the switch statement in C++, it seems like it only works with enums, which is also behaving a bit strangely in C++ compared to C#. I might do some refactoring at some point where I explore enums and switch statements and change all the large if clusters to switch statements.

utils.h
MortenTobiasNielsen: Well I started with the once I knew I would need to a short run (steel axe, GOTLAP). That turned out to not be as painful as I would have imagined, so I took a little more and in the end ended up with most of it. :)

utils.h
MortenTobiasNielsen: Yeah it is sort of strange to do it like this and I would also like the enums from C#. I do though also need the actual text string name of these most places, so I would have to do some kind of "ToString" all over the place and I am not sure how enums behave in C++.

MortenTobiasNielsen (commented):
Answering comments

cMain.cpp
MortenTobiasNielsen: This is to check that the building being build is a stone/steel furnace and make a put coal task if that is the case. It is not to put stone/steel/copper/iron into a furnace. :)

cMain.cpp
```diff
@@ -2141,6 +2141,40 @@ void cMain::OnMenuOpen(wxCommandEvent& event) {
 					malformed_saved_file_message();
 					return;
 				}
+			} else if (!auto_put_reached) {
+				if (seglist.size() == 2) {
+					if (seglist[0] == auto_put_furnace_text) {
+						if (seglist[1] == "true") {
+							check_furnace->SetValue(true);
+						} else {
+							check_furnace->SetValue(false);
```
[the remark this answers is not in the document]
MortenTobiasNielsen: That is so much cleaner - thanks.

cMain.cpp
MortenTobiasNielsen: I have made an issue to "Refactor away from if clusters"

utils.h
MortenTobiasNielsen: I have made an issue to "Refactor and use utils.h objects".

cMain.cpp
theis999: Oh i see, that makes sense.

cMain.cpp
theis999: If it works, then it is alright.

utils.h
theis999: A reasonable approach 😉

utils.h
theis999: One alternative to structs, is to use namespaces

`namespace science_packs{
 std::string red = "automation-science-pack", green = "logistic-science-pack"
}`

theis999 (approved):

utils.h
MortenTobiasNielsen: I see I will have that in mind if I am the one taking the refactor task.

PR #46, opened 2022-09-09
Added auto-close settings to save file

MortenTobiasNielsen:
I thought that it made sense to add auto-put settings, given that it is very related to this change, but the auto-put changes are not on this branch. So I add this branch to the auto-put branch and add the save logic there instead.

### DIFF
--- a/cMain.cpp
+++ b/cMain.cpp
@@ -3897,10 +3897,10 @@ bool cMain::save_file(bool save_as) {
 	}
 
 	myfile << auto_close_indicator << std::endl;
-	myfile << auto_close_generate_script_text + ";" + bool_to_string(menu_auto_close->GetMenuItems()[0]->IsChecked()) << std::endl;
-	myfile << auto_close_open_text + ";" + bool_to_string(menu_auto_close->GetMenuItems()[1]->IsChecked()) << std::endl;
-	myfile << auto_close_save_text + ";" + bool_to_string(menu_auto_close->GetMenuItems()[2]->IsChecked()) << std::endl;
-	myfile << auto_close_save_as_text + ";" + bool_to_string(menu_auto_close->GetMenuItems()[3]->IsChecked());
+	myfile << auto_close_generate_script_text << ";" << bool_to_string(menu_auto_close->GetMenuItems()[0]->IsChecked()) << std::endl;
+	myfile << auto_close_open_text << ";" << bool_to_string(menu_auto_close->GetMenuItems()[1]->IsChecked()) << std::endl;
+	myfile << auto_close_save_text << ";" << bool_to_string(menu_auto_close->GetMenuItems()[2]->IsChecked()) << std::endl;
+	myfile << auto_close_save_as_text << ";" << bool_to_string(menu_auto_close->GetMenuItems()[3]->IsChecked());
 
 	myfile.close();
 
--- a/cMain.cpp
+++ b/cMain.cpp
@@ -1825,6 +1825,9 @@ void cMain::OnMenuOpen(wxCommandEvent& event) {
 		bool templates_reached = false;
 		bool save_file_reached = false;
 		bool task_file_reached = false;
+		bool generate_code_folder_reached = false;
+		bool auto_close_reached = false;
+		bool auto_put_reached = false;
 
 		bool groups_in_file = false;
 		bool templates_in_file = false;
@@ -1868,8 +1871,7 @@ void cMain::OnMenuOpen(wxCommandEvent& event) {
 				if (seglist[0] == total_tasks_indicator) {
 					total_tasks_reached = true;
 				} else {
-					reset_to_new_window();
-					wxMessageBox("It seems like the structure of the file does not correspond with an EZRaiderz TAS helper file", "A file error occurred");
+					malformed_saved_file_message();
 					return;
 				}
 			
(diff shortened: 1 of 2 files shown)


cMain.cpp
```diff
@@ -1868,8 +1871,7 @@ void cMain::OnMenuOpen(wxCommandEvent& event) {
 				if (seglist[0] == total_tasks_indicator) {
 					total_tasks_reached = true;
 				} else {
-					reset_to_new_window();
-					wxMessageBox("It seems like the structure of the file does not correspond with an EZRaiderz TAS helper file", "A file error occurred");
+					malformed_saved_file_message();
```
theis999, line 1874: Moving this to function seem like a good change.

cMain.cpp
```diff
@@ -1896,21 +1898,35 @@ void cMain::OnMenuOpen(wxCommandEvent& event) {
...
+				task_file_reached = true;
+				generate_code_folder_reached = true;
+
+			} else if (seglist[0] == auto_put_indicator) {
+				groups_reached = true;
+				templates_reached = true;
+				save_file_reached = true;
+				task_file_reached = true;
+				generate_code_folder_reached = true;
+				auto_close_reached = true;
```
theis999, line 1916: It looks like there is a lot of code overlap here, which could mean there is better way to do this.

cMain.cpp
```diff
@@ -2048,16 +2060,61 @@ void cMain::OnMenuOpen(wxCommandEvent& event) {
...
 
-			} else {
+			} else if (!generate_code_folder_reached) {
 				generate_code_folder_location = seglist[0];
+
+			} else if (!auto_close_reached) {
+				if (seglist.size() == 2) {
+					if (seglist[0] == auto_close_generate_script_text) {
+						if (seglist[1] == "true") {
+							menu_auto_close->GetMenuItems()[0]->Check();
```
theis999, line 2077: There are 4 identical if/else blocks, with the only difference being menu_auto_close->GetMenuItems()**[0]**. We can enumerate auto_close_generate_script_text=0, auto_close_open_text=1... etc. so we only need one if/else block

cMain.cpp
```diff
@@ -3878,12 +3890,18 @@ bool cMain::save_file(bool save_as) {
...
 		myfile << save_file_indicator << std::endl;
-		myfile << save_file_location;
+		myfile << save_file_location << std::endl;
 	}
 
+	myfile << auto_close_indicator << std::endl;
+	myfile << auto_close_generate_script_text + ";" + bool_to_string(menu_auto_close->GetMenuItems()[0]->IsChecked()) << std::endl;
+	myfile << auto_close_open_text + ";" + bool_to_string(menu_auto_close->GetMenuItems()[1]->IsChecked()) << std::endl;
+	myfile << auto_close_save_text + ";" + bool_to_string(menu_auto_close->GetMenuItems()[2]->IsChecked()) << std::endl;
+	myfile << auto_close_save_as_text + ";" + bool_to_string(menu_auto_close->GetMenuItems()[3]->IsChecked());
```
theis999, line 3903: It feels weird not having std::endl here.

cMain.cpp
```diff
@@ -3878,12 +3890,18 @@ bool cMain::save_file(bool save_as) {
...
-		myfile << generate_code_folder_location;
+		myfile << generate_code_folder_location << std::endl;
 	} else {
 		myfile << save_file_indicator << std::endl;
-		myfile << save_file_location;
+		myfile << save_file_location << std::endl;
 	}
 
+	myfile << auto_close_indicator << std::endl;
+	myfile << auto_close_generate_script_text + ";" + bool_to_string(menu_auto_close->GetMenuItems()[0]->IsChecked()) << std::endl;
```
theis999, line 3900: It seems odd to mix << and string concatenation this way. An alternative is: 
`myfile << auto_close_generate_script_text << ";" << bool_to_string(menu_auto_close->GetMenuItems()[0]->IsChecked()) << std::endl;`

theis999 (commented):
I don't love the **if (!auto_close_reached)** segment. otherwise the code seem solid.

cMain.cpp
MortenTobiasNielsen: Yeah it seems like there should be a better way to do this, but I haven't found one. The problem is that all the sections are in the same file, but all of them won't always be there, e.g. if there are no groups the groups section won't be there, so when I get to the template function I need to set both groups and template as being reached.

I should probably have made it so all the sections would always be there.

cMain.cpp
MortenTobiasNielsen: I understand why you would want to do that, it would be much more concise. I do though think it should be part of a larger rewrite of the entire "OnMenuOpen" and "OnMenuSave" functions.

cMain.cpp
MortenTobiasNielsen: Agreed that makes more sense.

cMain.cpp
MortenTobiasNielsen: It is the last line of the file. I am not sure how the open function would like an empty line at the bottom.

cMain.cpp
MortenTobiasNielsen: I have made an issue to fix the "OnMenuOpen" and "OnMenuSave" functions.

cMain.cpp
MortenTobiasNielsen: I have made an issue to fix the "OnMenuOpen" and "OnMenuSave" functions.

cMain.cpp
MortenTobiasNielsen: I have made an issue to fix the "OnMenuOpen" and "OnMenuSave" functions.

PR #50, opened 2022-09-10
Minor changes based on testing

MortenTobiasNielsen:
Also changed the default quick bar, so it aligns with what I think most runs would see the most.


theis999 (approved):

theis999:
It looks fine

PR #51, opened 2022-09-10
Feature/move save task to not consume tick

MortenTobiasNielsen:
I also removed the debug massage writing what was in a rock - I don't understand what I as a TAS developer would use it for.


theis999 (approved):

theis999:
It looks fine.

PR #53, opened 2022-09-10
Convert lua strings to files

theis999:
### Just for show

Lua code is moved to separate files. Those files are set to output on build.

### File settings
The files will be moved from the TAS-helper to the TAS mod on script generation. Script files that up to date is skipped, except the goal file and the step file.
The goal file changes based on which goal you choose, but the copy_from file remains unchanged which is why it needs to be updated.

### Lua setup
The goals are split into separate files. 
control.lua now require a goal.lua file
control1 and control2 is merged together


MortenTobiasNielsen:
Is this up to date with the master branch?

theis999:
It was as of 2 hours ago

info.json
```diff
@@ -0,0 +1,9 @@
+{
+	"name": "test_1.0",
+	"version": "0.0.2",
```
MortenTobiasNielsen, line 3: This should be version 0.0.5

MortenTobiasNielsen (changes requested):

MortenTobiasNielsen:
control_info.h should be cleaned or deleted.

info.json
theis999: Correct. Looking at it, info.json can't be made this way. Since we need to change the mod name.

MortenTobiasNielsen: Ohh yeah you are right. Info might live best in c++. :)

theis999: Agreed, i reverted to the old way.

theis999:
I have reverted to the old info.json and cleaned the control_info.h

PR #54, opened 2022-09-10
Fixed a bug when deleting tasks.

MortenTobiasNielsen:


theis999 (approved):
I haven't looked at the delete code before and i honestly have no idea how it works.
But this change looks slightly cleaner

MortenTobiasNielsen:
Understandable it is also one of the more/most convoluted parts of the code base.

PR #55, opened 2022-09-11
Moved Buildings tab to the third tab on the lidt

MortenTobiasNielsen:
During my testing I found it a little frustrating to have buildings selected by default, so I have moved it to the third tab instead.


theis999 (approved):
Since it is just generated files i'll just approve it.

PR #56, opened 2022-09-11
Optimize script generation time

MortenTobiasNielsen:
I have left a few comments, to make it a little easier to understand the changes.

### DIFF
--- a/cMain.cpp
+++ b/cMain.cpp
@@ -3517,44 +3517,48 @@ void cMain::find_new_orientation() {
 
 bool cMain::find_building_for_script(int& row) {
 	for (int i = row - 1; i > -1; i--) {
-		if (grid_tasks->GetCellValue(i, 0) == "Build") {
-			building_x_cord = grid_tasks->GetCellValue(i, 1);
-			building_y_cord = grid_tasks->GetCellValue(i, 2);
+		split_task(tasks_data_to_save[i]);
+
+		if (task_segments[0] == struct_tasks_list.build) {
+			building_x_cord = task_segments[1];
+			building_y_cord = task_segments[2];
 
 			if (x_cord == building_x_cord && y_cord == building_y_cord) {
-				building = grid_tasks->GetCellValue(i, 4).ToStdString();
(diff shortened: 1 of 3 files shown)


cMain.cpp
```diff
@@ -3489,51 +3517,70 @@ void cMain::find_new_orientation() {
 
 bool cMain::find_building_for_script(int& row) {
 	for (int i = row - 1; i > -1; i--) {
-		building_x_cord = grid_tasks->GetCellValue(i, 1);
-		building_y_cord = grid_tasks->GetCellValue(i, 2);
```
MortenTobiasNielsen, line 3493: It turns out that it was primarily due to these lines being moved out of the if statements. I should have thought about that - this method is executed 1000's of times on script generation, so to get the x and y coordinates when they are not needed is just bad practice. :)

cMain.cpp
```diff
@@ -3489,51 +3517,70 @@ void cMain::find_new_orientation() {
...
+		if (compare_task_strings(grid_tasks->GetCellValue(i, 0), struct_tasks_list.build)) {
```
MortenTobiasNielsen, line 3520: Apparently the string comparer isn't that efficient. This turns out to be much better)

utils.h
```diff
@@ -785,7 +785,7 @@ struct {
 	std::string uranium_fuel_cell = "Uranium fuel cell";
 } struct_fuel_list;
 
-struct {
+static const struct {
```
MortenTobiasNielsen, line 788: I don't think turning these into static const structs had much to do with the performance improvements, but might as well have them.

cMain.cpp
theis999: It makes sense that calls to the gui would be the culprit

theis999 (approved):
It looks reasonable.
If it continues being a problem then we could look at either multithreading the process or make a prepared output.

MortenTobiasNielsen:
Yeah I was also thinking about multithreading, but it it can also get extremely convoluted.

PR #60, opened 2022-09-11
Optimize move task

MortenTobiasNielsen:
It is a little faster but not a fast as I would have hoped. I think we need an in-memory representation of the buildings grid (like task_data_to_save) to be able to optimize it further.

### DIFF
--- a/cMain.cpp
+++ b/cMain.cpp
@@ -776,7 +776,8 @@ void cMain::update_buildings() {
 
 	update_buildings_grid();
 
-	for (int i = 1; i < std::stoi(building_amount_of_buildings); i++) {
+	int amount_of_buildings = std::stoi(building_amount_of_buildings);
+	for (int i = 1; i < amount_of_buildings; i++) {
 		find_coordinates(building_x_cord, building_y_cord, building_direction_to_build, building_building_size);
 		update_buildings_grid();
 	}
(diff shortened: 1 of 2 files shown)


cMain.cpp
```diff
@@ -3832,8 +3840,9 @@ bool cMain::check_buildings_grid() {
 		}
 
 	} else if (building_task == "Take" || building_task == "Put") {
-		for (int i = 0; i < std::stoi(building_amount_of_buildings); i++) {
-			if (from_into != "Wreck" && !find_building()) {
+		building_amount_of_buildings_int = std::stoi(building_amount_of_buildings);
+		for (int i = 0; i < building_amount_of_buildings_int; i++) {
+			if (from_into != "Wreck" && !find_building(building_amount_of_buildings_int)) {
 				wxMessageBox("Building location doesn't exist.\n1. Please use exactly the same coordinates as you used to build \n2. Check that you have not removed the building(s)\n3. Check that you are not putting this task before the Build task", "Please use the same coordinates");
```
theis999, line 3846: Didn't you add an error msg handler?

cMain.cpp
```diff
@@ -3842,8 +3851,7 @@ bool cMain::check_buildings_grid() {
 		}
 
 	} else if (building_task == "Launch") {
-		building_amount_of_buildings = "1";
-		if (!find_building()) {
+		if (!find_building(1)) {
 			wxMessageBox("Building location doesn't exist.\n1. Please use exactly the same coordinates as you used to build \n2. Check that you have not removed the building(s)\n3. Check that you are not putting this task before the Build task", "Please use the same coordinates");
```
theis999, line 3855: Didn't you add an error msg handler?

theis999 (approved):
👍

cMain.cpp
MortenTobiasNielsen: What do you mean?

cMain.cpp
MortenTobiasNielsen: What do you mean?

theis999: For "wxMessageBox" i seem to recall you made a function to handle showing the error box. Maybe that was only related to the save file?

MortenTobiasNielsen: Ohh yeah, you are right, but it was specific to save file.

PR #61, opened 2022-09-11
Optimized open file time

MortenTobiasNielsen:


MortenTobiasNielsen:
Leave this until PR 60 has been merged - I have build this on top, which I shouldn't have.

cMain.cpp
```diff
@@ -2219,6 +2223,38 @@ void cMain::OnMenuOpen(wxCommandEvent& event) {
...
+void cMain::populate_tasks_grid() {
+	if (grid_tasks->GetNumberRows() > 0) {
+		grid_tasks->DeleteRows(0, grid_tasks->GetNumberRows());
+	}
+
+	size_t amount_of_tasks = tasks_data_to_save.size();
+
+	grid_tasks->InsertRows(0, amount_of_tasks);
+
+	for (unsigned int i = 0; i < amount_of_tasks; i++) {
```
theis999, line 2235: does it need to be an unsigned int, that seems unusual

theis999 (approved):

cMain.cpp
MortenTobiasNielsen: It doesn't need to be unsigned, I just pasted it from somewhere else in the code, where I apparently though it made sense given that I can never be negative. :)

PR #62, opened 2022-09-11
Added a line so it is always day

### LINKED ISSUE #59: Always day
`/c game.player.surface.always_day=true`
For viewing purpose it would be nice have consistent lighting. 
I think we can add it to the new settings file.

MortenTobiasNielsen:
closes #59

### DIFF
--- a/control_info.h
+++ b/control_info.h
@@ -702,6 +702,7 @@ script.on_event(defines.events.on_tick, function(event)
 	if not run then return end
     if not player then 
 		player = game.players[1]
+		player.surface.always_day=true
 		player_position = player.position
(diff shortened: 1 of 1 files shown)


control_info.h
```diff
@@ -702,6 +702,7 @@ script.on_event(defines.events.on_tick, function(event)
 	if not run then return end
     if not player then 
 		player = game.players[1]
+		player.surface.always_day=true
```
theis999, line 705: I am not confident i want this to run at every tick

MortenTobiasNielsen: This will only run once at the very start when the player is not set.

theis999: Ah perfect!

PR #63, opened 2022-09-11
Solved an issue where comments wouldn't be loaded from save file

MortenTobiasNielsen:


cMain.cpp
```diff
@@ -2040,7 +2046,7 @@ void cMain::OnMenuOpen(wxCommandEvent& event) {
 			} else if (!templates_reached) {
 				groups_in_file = true;
 
-				if (seglist.size() == 10) {
+				if (seglist.size() == 10 || seglist.size() == 11) {
```
theis999, line 2049: What happens at segment 11?

theis999 (approved):

cMain.cpp
MortenTobiasNielsen: Segment 11 is a comment, which won't be in old saved files, so it is needed for backwards compatibility.

theis999: Fair

PR #65, opened 2022-09-13
Skip cutscene

theis999:
Skips cutscene and creation of the crashsite
Then recreates the crashsite

### DIFF
--- a/control.lua
+++ b/control.lua
@@ -1,4 +1,7 @@
 require("goal")
+local util = require("util")
+local crash_site = require("crash-site")
+
 local steps = require("steps")
 local debug_state = false
 local run = true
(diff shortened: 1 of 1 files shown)


MortenTobiasNielsen:
@theis999 please merge master into this branch, so I a little easier can understand what has actually happened. :)

control.lua
```diff
@@ -1,4 +1,7 @@
 require("goal")
+local util = require("util")
+local crash_site = require("crash-site")
```
MortenTobiasNielsen, line 3: Just for my understanding - where does this come form?

control.lua
```diff
@@ -37,6 +40,23 @@ local neg_neg = false
...
+	local surface = player.surface
+	local crashed_ship_items = remote.call("freeplay", "get_ship_items")
+	local crashed_debris_items = remote.call("freeplay", "get_debris_items")
+	util.insert_safe(player, global.created_items)
+
+    surface.daytime = 0.7
+    crash_site.create_crash_site(surface, {-5,-6}, util.copy(crashed_ship_items), util.copy(crashed_debris_items))
+    util.remove_safe(player, crashed_ship_items)
+    util.remove_safe(player, crashed_debris_items)
+    player.get_main_inventory().sort_and_merge()
```
MortenTobiasNielsen, line 57: How will this affect players inventory if it is a saved game? It seems like it will change how the player has arranged their inventory.

control.lua
theis999: [util](https://github.com/wube/factorio-data/blob/master/core/lualib/util.lua)
[crash_site](https://github.com/wube/factorio-data/blob/master/core/lualib/crash-site.lua)

They are part of factorio

control.lua
theis999: It is only **on player created**, which should only happen once throughout a game.
It might not be mp safe, but returning to save games is fine.

PR #67, opened 2022-09-14
Rock setting

theis999:
Added a startup setting for forced rock yields


MortenTobiasNielsen:
I am not sure what I think about this. The hope of having "cheaty_rocks" is that everyone doing TAS will be using it either by using the tool or by setting it by themselves, so TAS runs are comparable.

MortenTobiasNielsen:
Fair, but why make it a setting in the first place instead of just having it hard coded?

MortenTobiasNielsen:
I don't think this makes sense to merge - I therefore close the pull request

PR #68, opened 2022-09-14
Changed name of Long handed inserter to Long-handed inserter, so it is aligned with the name in game

MortenTobiasNielsen:


cMain.cpp
```diff
@@ -2332,6 +2337,11 @@ void cMain::OnGenerateScript(wxCommandEvent& event) {
 	size_t amount_of_tasks = tasks_data_to_save.size();
 
 	for (int i = 0; i < amount_of_tasks; i++) {
+		
+		if (i > 2937) {
+			int test = 1;
```
theis999, line 2342: This looks suspect. Is it some sort of debug code?

cMain.cpp
```diff
@@ -2520,49 +2530,65 @@ void cMain::OnGenerateScript(wxCommandEvent& event) {
...
-	saver.open(generate_code_folder_location + "\\steps.lua");
+			saver.close();
+
+			saver.open(generate_code_folder_location + "\\info.json");
+			saver << "\{\n\t\"name\": \"" + generate_code_folder_location.substr(generate_code_folder_location.rfind("\\") + 1) + "\",";
+			saver << "\n\t\"version\": \"" << software_version << "\",";
+			saver << "\n\t\"title\": \"" + generate_code_folder_location.substr(generate_code_folder_location.rfind("\\") + 1) + "\",";
+			saver << info;
+			saver.close();
+		}
```
theis999, line 2569: I don't understand what you are trying to achieve here?

cMain.cpp
MortenTobiasNielsen: Yeah it is. :D

cMain.cpp
MortenTobiasNielsen: "// If the file is send to another person the folder location won't exist and should be set to something else."

MortenTobiasNielsen:
I merge master in and try to solve the problem again, so it works with the current code.

theis999 (approved):
Looks good

PR #70, opened 2022-09-15
Compiled the full recipe list

theis999:
Updated the full list of recipes with the help of lua code.


MortenTobiasNielsen:
Very nice - we will need to ensure that the added recipes are also available as buildings. It might be a little frustration that some things cannot be crafted, but at least it is possible to place everything which can be crafted. :D

theis999:
Looking over it again i can see i am missing comma between ingredient items, so i will have to fix that.
I can also see that the list includes loaders and rocket parts, so i will remove those.

Does that cover your concerns?

MortenTobiasNielsen:
I think I would like to save this for the C# version, I would rather spend the mental capacity testing it in that context. :)

theis999:
@MortenTobiasNielsen I agree

MortenTobiasNielsen:
> Looking over it again i can see i am missing comma between ingredient items, so i will have to fix that. I can also see that the list includes loaders and rocket parts, so i will remove those.
> 
> Does that cover your concerns?

If you fix those problems, then I will try to merge it ins and see if it causes problems. :)

theis999:
@MortenTobiasNielsen done!

src/utils.h
```diff
@@ -78,62 +78,216 @@ static const std::vector<std::string> splitter_list = { "Splitter",
...
-																		 { "Logistic science pack", {"Inserter", "1", "Transport belt", "1" }},
-																		 { "Chemical science pack", {"Advanced circuit", "3", "Engine unit", "2", "Sulfur", "1" }},
-																		 { "Production science pack", {"Electric furnace", "1", "Productivity module", "1", "Rail", "30" }},
-																		 { "Utility science pack", {"Flying robot frame", "1", "Low density structure", "3", "Processing unit", "2" }},
-																	   };
+static const std::map<std::string, std::vector<std::string>> recipes = { 
+	{ "Assembling machine 1", { "Iron plate", "9" , "Iron gear wheel", "5" , "Electronic circuit", "3"  }},
+	{ "Automation science pack", { "Copper plate", "1" , "Iron gear wheel", "1"  }},
+	{ "Firearm magazine", { "Iron plate", "4"  }},
+	{ "Pistol", { "Iron plate", "5" , "Copper plate", "5"  }},
```
MortenTobiasNielsen, line 85: Sorry for the late reply. There is a space after the amount, which I don't think should be there, so it is consistent with the spacing of the name.

src/utils.h
```diff
@@ -78,62 +78,216 @@ static const std::vector<std::string> splitter_list = { "Splitter",
...
+	{ "Small lamp", { "Iron plate", "1" , "Copper cable", "3" , "Electronic circuit", "1"  }},
+	{ "Basic oil processing", { "Crude oil", "100"  }},
+	{ "Advanced oil processing", { "Water", "50" , "Crude oil", "100"  }},
+	{ "Coal liquefaction", { "Coal", "10" , "Heavy oil", "25" , "Steam", "50"  }},
+	{ "Pipe", { "Iron plate", "1"  }},
+	{ "Pipe to ground", { "Iron plate", "5" , "Pipe", "10"  }},
+	{ "Logistic robot", { "Advanced circuit", "2" , "Flying robot frame", "1"  }},
+	{ "Construction robot", { "Electronic circuit", "2" , "Flying robot frame", "1"  }},
+	{ "Energy shield equipment", { "Steel plate", "10" , "Advanced circuit", "5"  }},
+	{ "Energy shield mk2 equipment", { "Processing unit", "5" , "Low density structure", "5" , "Energy shield equipment", "10"  }},
```
MortenTobiasNielsen, line 117: I think all of the "equipment" need to have a conversion, The name in the game is Energy shield MK2 and I have made the item list like the once in the game, so when I chose to set the recipe for Energy shield MK2 it isn't able to find the recipe. You might be able to use the "GenerateScript::check_item_name" method for the conversion.

MortenTobiasNielsen (commented):
Looks good, but I have two comments.

src/utils.h
theis999: I ran it through a format tool, so it should be alright now.

src/utils.h
theis999: [Energy_shield_MK2](https://wiki.factorio.com/Energy_shield_MK2)
The external english name is `Energy shield MK2` but the internal name is `energy-shield-mk2-equipment` both of the item and the recipe (and technology). 
If we need to make many conversions, then we should have a conversion table.

MortenTobiasNielsen: I haven't run into anything else than the equipment

theis999:
Changed the names of equipments

MortenTobiasNielsen:
I am pretty sure that won't work - have you tested it?

theis999:
It does work, in my tests

MortenTobiasNielsen:
I get this error:
image

theis999:
I don't understand why mk2 should be lowercase, but ok

MortenTobiasNielsen:
I will merge it in and make the changes on my testing branch.

PR #71, opened 2022-09-15
Updated the building size list

theis999:
Updated the building size list with the help of lua.

I think we need to revisit how it is used, to make it accurate.


theis999:
i didn't include [secondary_collision_box](https://lua-api.factorio.com/latest/LuaEntityPrototype.html#LuaEntityPrototype.secondary_collision_box)

The output named "Active provider chest" => "Logistic chest active provider". I am not sure what is correct, so i used the name you used.

I cleaned out shipwreck parts (among other things), but the wreck collition boxes could still be useful somewhere.

theis999:
@MortenTobiasNielsen 
I have reworked the check_interact_distance distance algorithm to be correct and so it works with the correct building sizes.

It now no longer uses the buffer distance. It now forces buildings to be built at either even numbers or offset by 0.5 if the building size is an uneven number of tiles.

I am not sure if this is the correct approach but it does appear to work quite well.

src/GenerateScript.cpp
```diff
@@ -508,25 +508,29 @@ void GenerateScript::check_interact_distance(std::string step, std::string actio
...
-	static const float max_distance = 10.0f;
-	float min_x_edge;
-	float max_x_edge;
+	static const float buffer = 0.0f; // TODO remove
+	static const float max_distance = 10.0f; //Default build distance
 
-	float min_y_edge;
-	float max_y_edge;
+	float x_target = floor(std::stof(x_cord));
+	if ((int)ceil(x_building_size) % 2 == 1) x_target += 0.5; //if a building is an uneven number of tiles wide, it will be placed at a half tile
```
MortenTobiasNielsen, line 515: 1. It seems like we end up with the same as the old list, if we ceil the numbers from the building_size_list.
2. I always specify the half tile when I place buildings with an uneven number. I therefore don't really like this approach of assuming that the user won't enter what they need to. 
3. I would like to have bracers here, even though it is a one liner - I find it a little hard to read without them.

src/GenerateScript.cpp
```diff
@@ -508,25 +508,29 @@ void GenerateScript::check_interact_distance(std::string step, std::string actio
...
+
+	float y_target = floor(std::stof(y_cord));
+	if ((int)ceil(y_building_size) % 2 == 1) y_target += 0.5; //if a building is an uneven number of tiles tall, it will be placed at a half tile
+
+	float min_x_edge = x_target, max_x_edge = x_target;
+	float min_y_edge = y_target, max_y_edge = y_target;
 
 	if (orientation == "North" || orientation == "South") {
```
MortenTobiasNielsen, line 523: I think this is some legacy code I have forgotten to get rid of, but just to be certain - The two blocks are completely the same, aren't they?

MortenTobiasNielsen:
This will completely de-sync a run, so we will need to have a way to specify that the old way should be used. I have done this for the running algorithm by allowing the user to specify "old" in the comment, maybe you could do the same?

src/GenerateScript.cpp
theis999: They are not the same, they swap the x and y on the right side of the equation. Which I believe is the correct approach.

src/GenerateScript.cpp
theis999: 1. The ceil of the old list and the ceil of the new list should be equal yes. But the ceil of the new list (eg. stone furnace {1.398438f, 1.398438f} => {2, 2}) is not the same as the old list (stone furnace {1.6f, 2}).
2. The algorithm does need the value to be correct, whether the user sets the building correctly or we correct it for them, is a matter of taste. Would you rather have this input correction another place or just not at all? 
3. Understandable, looking at it again it is quite dense.

MortenTobiasNielsen: 1. True - good that it is correct.
2. It has always needed to be correct and I haven't had anyone complaining about it being difficult to use, so I rather not have it being corrected.

theis999: I have left it commented out, then we can revisit the idea.

theis999:
Rebased on master:

Removed input correction
Re-added old list, added a pointer to the building list
set pointer to either the new list or old list based on comment "old" / "new"

src/GenerateScript.cpp
```diff
@@ -512,34 +512,51 @@ void GenerateScript::check_mining_distance(std::string step, std::string action,
 }
 
 void GenerateScript::check_interact_distance(std::string step, std::string action, std::string x_cord, std::string y_cord, std::string building_name, std::string orientation) {
+	//if comment is "old" then use old map and buffer = 0.37 until a comment of new
+	if (comment == "old" || comment == "Old") {// TODO remove
+		building_size_map_p = &old_building_size_list; 
+	}
+	else if (comment == "new" || comment == "New") {
```
MortenTobiasNielsen, line 519: Can I maybe persuade you to implement a separation of the two? Purely for my own selfish gain. I have a lot of tasks with the new walking algorithm, which would be affected by this, but I do want to use it (and test it). I know it won't affect the users given that they will get the full benefit - unless someone have compiled from source.

MortenTobiasNielsen (commented):
I would like to be able to test it, but I won't be able to if the two changes to walking use the same identifier. :(

src/GenerateScript.cpp
```diff
@@ -512,34 +512,51 @@ void GenerateScript::check_mining_distance(std::string step, std::string action,
...
+	if ((int)ceil(x_building_size) % 2 == 1) x_target += 0.5; //if a building is an uneven number of tiles wide, it will be placed at a half tile
+
+	float y_target = floor(std::stof(y_cord));
+	if ((int)ceil(y_building_size) % 2 == 1) y_target += 0.5; //if a building is an uneven number of tiles tall, it will be placed at a half tile
+	*/
 
-	float min_y_edge;
-	float max_y_edge;
+	float min_x_edge = x_target, max_x_edge = x_target;
+	float min_y_edge = y_target, max_y_edge = y_target;
```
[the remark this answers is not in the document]
theis999: Why not?

src/GenerateScript.cpp
theis999: I added extra comment options, is that what you asked for?

src/GenerateScript.cpp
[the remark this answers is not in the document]
MortenTobiasNielsen: Mostly that I am not used to seeing them and therefore isn't as easy to read as just having them on their own lines.

src/GenerateScript.cpp
MortenTobiasNielsen: Somewhat - won't I need to put a comment on each build step?

theis999: No, just once per change. It toggles on a comment of new/old on any building interaction task, by setting the pointer to the new/old list. and determines if the buffer should be set to 0/old value based on where the pointer is pointing.

src/GenerateScript.cpp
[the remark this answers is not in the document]
theis999: I will make an effort of making you more used to it then :p

Issue #47, opened 2022-09-10
Refactor away from if clusters


MortenTobiasNielsen:
There are multiple places in the codebase where there are massive clusters of if statements. Evaluate how this can be done better and implement the change.

MortenTobiasNielsen:
I think so when we get the current branch merged in.

PR #74, opened 2022-09-18
Refactor switch statement

theis999:
#47 Refactor switch statement in cMain,cpp
The code is mostly the same except it uses switch blocks instead if blocks.
I added the comment _//fallthrough_ every time it intentionally uses more than one case for the code block.
I added Ternary operators a few places when it made sense.

### DIFF
--- a/cMain.cpp
+++ b/cMain.cpp
@@ -2341,8 +2341,8 @@ void cMain::OnGenerateScript(wxCommandEvent& event) {
 
 	for (int i = 0; i < amount_of_tasks; i++) {
 		In_memory_extract_parameters(tasks_data_to_save[i]);
-
-		if (task == "Start") {
+		auto t = map_task_name[task];
+		if (t == e_start) {
 			clear_tasks();
 		}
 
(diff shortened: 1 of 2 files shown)


cMain.cpp
```diff
@@ -592,32 +585,27 @@ bool cMain::change_row(wxGrid* grid) {
...
-	if (task == "Build") {
-		grid->SetCellBackgroundColour(row, 0, *wxCYAN);
-		return;
-	}
-
-	if (task == "Craft") {
-		grid->SetCellBackgroundColour(row, 0, *wxLIGHT_GREY);
-		return;
+	switch (map_task_name[task]) {
+		case e_start:
```
MortenTobiasNielsen, line 589: Is the e to indicate an enum?

cMain.h
```diff
@@ -9,8 +9,17 @@
...
 
+#include <iomanip>
+#include <locale>
+#include <codecvt>
+#include <filesystem>
+
+
 #include "GUI_Base.h"
 #include "ScriptProgressBar.h"
+#include "icon.xpm"
```
MortenTobiasNielsen, line 20: What is this used for?

MortenTobiasNielsen:
@theis999 You once mentioned that you would like to have C# enums - what do you think about porting the entire project to C# and use .Net MAUI for the front-end? At work I use C#, so it would be more beneficial for me to learn about C# optimization and I would have better knowledge in terms of structuring the project for easier extend-ability.

cMain.cpp
theis999: Yes, and the compiler couldn't handle that the enum was called `start` while there is also a function called `start`.
I am open for other identifiers if it is a problem

MortenTobiasNielsen: A little strange that it cannot distinguish between the two. :) It seems like a good way to solve the issue.

cMain.h
theis999: I believe it is used to display the ez-raiders icon. It is used among other places in the cmain constructor:


`cMain::cMain() : GUI_Base(nullptr, wxID_ANY, window_title, wxPoint(30, 30), wxSize(1840, 950)) {
	SetIcon(icon_xpm);`

MortenTobiasNielsen: I didn't see that it was just moved from .cpp to .h. :) I think you are right that it is the icon handling.

theis999:
@MortenTobiasNielsen I am not opposed to that.
I think C# is a much stronger language and [lua-in-c#](http://nlua.org/) looks slightly better than [lua-in-cpp](https://www.geeksforgeeks.org/integrating-lua-in-cpp/).

It seems like MAUI doesn't support linux, on the other hand it's styling options looks much better compared to wxForms.

MortenTobiasNielsen:
@theis999 It seems like .net6 is Linux compatible, so I guess it is only a matter of time before MAUI will support Linux. Yeah it would provide much better options for UI - like having more than 6 useable colors. :D

theis999:
If linux support is on the way, then the downside is that might be a bit of work.
But the improved styling options seems more than worth it, so i am all for it.
So we should wrap up a final release on the c++ platform and then look at a conversion to c#

Personally i would like to see the gui styled a bit more factorio styled

MortenTobiasNielsen:
The released versions are only for Windows computers - as far as I know, so unless we know that someone on Linux are compiling from source I don't think it is a problem that Linux isn't supported for some time. It will for sure be a good amount of work to do it correctly and then most likely even more work to get it to perform as well as the C++ version. 

Yeah, that would be really cool.

PR #75, opened 2022-09-19
Added remote interface to expose the task list

theis999:
Added remote interface that exposes:
current task number
complete task list
an id: for other mods to listen for changes

Changed all changes of **step** to use function _change_step_
_change_step_ raises an event for other mods to listen for

### DIFF
--- a/control.lua
+++ b/control.lua
@@ -40,10 +40,12 @@ local neg_neg = false
 local drop_item
 local drop_position
 
+local tas_step_change = script.generate_event_name()
+
 --recreate crash site
 local on_player_created = function(event)
 	if remote.interfaces["freeplay"] == nil then return end
-
+	
 	local player = game.get_player(event.player_index)
 	local surface = player.surface
 	local crashed_ship_items = remote.call("freeplay", "get_ship_items")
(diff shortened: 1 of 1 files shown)


control.lua
```diff
@@ -308,7 +320,7 @@ local function build()
 		elseif player.can_place_entity{name = item, position = target_position, direction = direction} then
 			if player.surface.can_fast_replace{name = item, position = target_position, direction = direction, force = "player"} then
 				if player.surface.create_entity{name = item, position = target_position, direction = direction, force="player", fast_replace=true, player=player, raise_built = true} then
-					step = step - 1
+					change_step(-1)
```
MortenTobiasNielsen, line 323: I have to admit that I cannot remember why -1 makes sense here - it seems really counter intuitive.

MortenTobiasNielsen (approved):

control.lua
theis999: It might be due to `fast_replace=true` but i am not sure.

MortenTobiasNielsen: That seems to be the case, I just don't really understand why the fast replace would "consume" the step, so it makes sense to go to the previous step again. Maybe I have programmed it so it double steps in this situation and it needs to be corrected.

PR #82, opened 2022-09-27
changes to the gui

theis999:
@MortenTobiasNielsen I have tried to move stuff around in the gui to make intuitive. 

Changed it so the gui is generated the correct place.
Moved group, template, building and task into aui notebook. Allowing the user more control of the interface.
Changed the grids to be modifiable allowing the user more control of the interface
Removed some empty spacers


MortenTobiasNielsen:
@theis999 I would like the tabs to be in the same place (Steps on the right and the rest on the left), mostly just because that is the way I am used to it. It really throws me when it is the other way around. :)

- I like the idea of being able to move steps more than one step, but I think it would be better with a spin input between the two buttons, which could then be used to determine how many lines the steps should be moved up/down. We do though need to ensure that the amount doesn't exceed range of the grid.
- I still don't really understand the Segments section - is that a finished feature?

theis999:
I moved steps to the right, turns out i should have used wxRIGHT instead of wxRight...

I am not confident that a spin+move is a good idea. I understand where you are going with it, i just don't like two-steps gui controls. It did however made me think about using right click to move 5 instead.

Segments are not completed yet. I currently have working everything working except eventhandlers of buttons.

MortenTobiasNielsen:
It is better to have 1 step for sure, but the user is pretty used to having to set different types of inputs to get the outcome they want, so I think it would be pretty intuitive. If they just want to move one row, they don't need to change it and if they want to move 100 they can set it to 100 instead of having to right click 20 times or left click 100 times. :)

It seems like Segment should have its own branch.

MortenTobiasNielsen:
That would get a little tiresome when I need to have 100 tasks. In that case I do have the possibility to use groups/templates, but I would find it easier to just mark the once I want to move and then moving them.

MortenTobiasNielsen:
I didn't notice that it is possible to make changes directly in the table, which is cool but it doesn't seem like the change is recorded. I would be very confused as a user if I am able to make changes but they aren't reflected in the generated code or in the saves I make. 

Resize the columns in the building tab, so they are able to be there without the horizontal scrollbar.

MortenTobiasNielsen:
I just spotted the reason I chose to use units instead of amount. It is used in the grids to separate amount of buildings (amount) and units (units). I don't see a good way to do this Amount of buildings are too long to be nicely in the grid.

theis999:
I think I got a little overeager on the possibilities. Editing the tables directly is seems to be to much work to get it to work before my refactor. So I'll remove it.
I will redo the move task 5 buttons.

theis999:
I noticed that move 5 performed terrible, due to move 1 performing bad. It seemed out of scope to fix it.

MortenTobiasNielsen:
UI changes:

- x and y should only have 1 decimal and I would like for them to be right aligned. Like it currently is on the master branch (I know it wasn't like that when you made these changes - sorry for the delay.)
- From/Into should be either From or into at initialization, now that it has been changed so it is either or when the application is used
- Number of buildings shouldn't be capped at 24. I might want to drag a transport belt 100's of tiles
- The upper part of the tabs (steps, groups, templates and buildings) should have the same height, so it doesn't jump around when going from one tap to the other. It will create some "dead" space, but I had made it like that on purpose.
- The grids are not properly aligned, so they either clip the columns when scroll bars are there or doesn't fill out the entire width (especially noticeable for buildings). 
- There shouldn't be a horizontal scroll bars.
- Amount is not supposed to go to -1 - the code will do that if the amount is set to 0 and it makes sense for the task. 
- The height of the upper part can be lowered a little - there is a little much space. 
image

src/cMain.cpp
```diff
@@ -1218,6 +1323,24 @@ void cMain::OnMoveDownClicked(wxCommandEvent& event) {
...
+
+// You have chosen to not make checks when tasks are moved, given that it most likely would make the function very clunky to use
+void cMain::OnMoveDownFiveClicked(wxMouseEvent& event) {
+	for (int i = 0; i < 5; i++){
+		move_row(grid_tasks, false);
+	}
+	
+	event.Skip();
+}
+
```
MortenTobiasNielsen, line 1343: Remove this code.

src/cMain.cpp
```diff
@@ -2284,18 +2408,17 @@ void cMain::OnAddMenuSelected(wxCommandEvent& event) {
 }
 
 void cMain::setup_paramters(std::vector<bool> parameters) {
-	spin_x_cord->Enable(parameters[0]);
-	spin_y_cord->Enable(parameters[1]);
-	spin_units->Enable(parameters[2]);
+	m_spin_x->Enable(parameters[0]);
+	m_spin_y->Enable(parameters[1]);
+	m_spin_amount->Enable(parameters[2]);
```
MortenTobiasNielsen, line 2413: why m_?

src/cMain.h
```diff
@@ -103,7 +103,9 @@ class cMain : public GUI_Base {
 	void OnChangeTaskClicked(wxCommandEvent& event);
 	void OnDeleteTaskClicked(wxCommandEvent& event);
 	void OnMoveUpClicked(wxCommandEvent& event);
-	void OnMoveDownClicked(wxCommandEvent& event); 
+	void OnMoveDownClicked(wxCommandEvent& event);
+	void OnMoveUpFiveClicked(wxMouseEvent& event);
+	void OnMoveDownFiveClicked(wxMouseEvent& event);
```
MortenTobiasNielsen, line 108: Remove these

src/cMain.cpp
```diff
@@ -152,6 +134,103 @@ cMain::cMain() : GUI_Base(nullptr, wxID_ANY, window_title, wxPoint(30, 30), wxSi
...
+	int start;
+	auto sel = grid_tasks->GetSelectedRows();
+	if (sel.empty()) start = 0;
+	else start = grid_tasks->GetSelectedRows().back() + 1;
+
+	for (int i = start; i < rows; i++) {
+		auto cell = grid_tasks->GetCellValue(i, 0).ToStdString();
+		if (cell.starts_with(str)) {
+			grid_tasks->SelectRow(i);
+			grid_tasks->ScrollLines(i-start);
```
MortenTobiasNielsen, line 178: It won't scroll all the way to the task if you have a lot of tasks - it stops 3-4 rows before, so it seems like the selection just disappeared. The problem gets very noticeable if you press enter multiple times.

src/cMain.cpp
```diff
@@ -152,6 +134,103 @@ cMain::cMain() : GUI_Base(nullptr, wxID_ANY, window_title, wxPoint(30, 30), wxSi
...
+			event.Skip();
+			return;
+		}
+		else continue;
+	}
+	event.Skip();
+}
+void cMain::BuildingSearchOnCancelButton(wxCommandEvent& event) {
+	event.Skip();// do nothing, it will clear the search box
 }
```
MortenTobiasNielsen, line 235: Remove new line

src/cMain.cpp
```diff
@@ -152,6 +134,103 @@ cMain::cMain() : GUI_Base(nullptr, wxID_ANY, window_title, wxPoint(30, 30), wxSi
...
+			grid_tasks->ScrollLines(id - rows[0]);
+			event.Skip();
+			return;
+		}
+		else continue;
+	}
+	TaskSeachOnSearchButton(event);
+	event.Skip();
+}
+void cMain::TaskSeachOnTextEnter(wxCommandEvent& event) {
```
MortenTobiasNielsen, line 159: Add new line between all these methods.

src/cMain.cpp
```diff
@@ -152,6 +134,103 @@ cMain::cMain() : GUI_Base(nullptr, wxID_ANY, window_title, wxPoint(30, 30), wxSi
...
+	event.Skip();// do nothing, it will clear the search box
+}
+void cMain::BuildingSearchOnText(wxCommandEvent& event) {
+	auto str = event.GetString();
+	auto rows = grid_tasks->GetSelectedRows();
+	for (auto id : rows) {
+		auto cell = grid_tasks->GetCellValue(id, 0).ToStdString();
+		if (cell.starts_with(str)) {
+			grid_tasks->SelectRow(id);
+			grid_tasks->ScrollLines(id - rows[0]);
```
MortenTobiasNielsen, line 196: The same problem as above. When you have 100's or even worse 1000's of tasks it get a little silly with the scrolling. It can take multiple seconds before it get to where it is supposed to be.

src/cMain.cpp
```diff
@@ -152,6 +134,103 @@ cMain::cMain() : GUI_Base(nullptr, wxID_ANY, window_title, wxPoint(30, 30), wxSi
...
+		return;
+	}
+	int start;
+	auto sel = grid_tasks->GetSelectedRows();
+	if (sel.empty()) start = 0;
+	else start = grid_tasks->GetSelectedRows().back() + 1;
+
+	for (int i = start; i < rows; i++) {
+		auto cell = grid_tasks->GetCellValue(i, 0).ToStdString();
+		if (cell.starts_with(str)) {
```
MortenTobiasNielsen, line 176: It is nice to be able to search, but when you have a lot of tasks and you want to search for that time you crafted 10 electric mining drills it isn't that helpful to only be able to search for Craft. I am not sure how to solve that problem though.

src/cMain.cpp
theis999: The auto suggested name was m_spin_xxyyy...

src/cMain.cpp
theis999: I didn't spend a lot of time on search, i just wanted something basic we can expand upon later. If that is alright with you?

src/cMain.cpp
theis999: I don't think that it is great issue that it takes some time, as long as the gui feels responsive then the user can wait. I guess we will get other problems when the search options are expanded and searching becomes slower.

src/cMain.cpp
theis999: Again, this is not supposed to be the end of implementing search. It works (reasonable) well, for the most basic operation which is to find your **start** task.

src/cMain.cpp
theis999: Why? what is wrong with it, besides it being slow.

src/cMain.h
theis999: Why? they work.

theis999:
### Changes we agree on:
I have set x & y to 1 decimal.
I have moved x, y, amount, build size & build amount to the right.
I have set label_from_into to start with the text` From:`.
I have set number of building cap to 250, i had forgotten about belts so 24 seemed like a reasonable high number.
I have added a spacer to the buildings tab and steps tab to align with the others.
I have set the cap of amount to [0 - 1,000,000].
I have remove the extra height on the top 3 panels. (a remnant from when there was a Tech field).
### Changes we might not agree on:
I have allowed all Pages in the book to expand when possible. Which will mean the right aligned buttons will move to right.

I can not control that the grids spawn scrollbars. It is not an option through the GUI builder. 

The GUI and the grids are much more dynamic than before, and the user can display all 4 pages at the same time or just 1. Each page can be adjusted in horizontal width or moved. Every column in any grid can be adjusted in size. 
This allows the user much more control to view what they find important, and it also means that it is much less critical that the GUI is perfectly adjusted at startup in my opinion.


Also, you seem to be confused about move 5, they react to right-click.

src/cMain.cpp
MortenTobiasNielsen: Is it use for anything?

src/cMain.cpp
MortenTobiasNielsen: I would like for it not to be there unless there is a very good reason to do it.

src/cMain.h
MortenTobiasNielsen: How do I trigger them?

src/cMain.cpp
MortenTobiasNielsen: I want to get to a point where I can make a release to address the bug you found.  I therefore don't want to merge something into the master branch which I know is working in a confusing way.

src/cMain.cpp
MortenTobiasNielsen: I don't want to wait multiple seconds when I am searching for something - what is the point in having a search function that is so slow that I could find it faster by just manually searching for it?

src/cMain.cpp
MortenTobiasNielsen: It would have been nice that the search functionality was kept in a branch by itself, if the intention was to make a proof of concept. If the only thing the search function is really good for is finding the start task then it would more more user friendly to have a "Find Start" button.

MortenTobiasNielsen:
That is fine. If the user chooses to expand the grid beyond the default width we cannot really do much about that with the GUI building - or at least I haven't found a way to make the entire grid responsive. 

If the column widths in the grid are configured to fit in the grid there won't be horizontal scroll bars. 

It is nice that the user can make some adjustments, but I would be surprised if a user will use any real amount of time customizing their view until those adjustments are saved. I would therefore like the default GUI to be setup in a way which looks nice - especially given that I won't be changing anything in the GUI - it is exactly like I would want it to be, it is almost like I designed it myself. :) 

I most certainly was confused about that. I didn't see any indication that I was able to do that in the GUI and I still don't understand why 5 - are there some common thing which needs to be done where a row needs to moved 5 rows?

src/cMain.cpp
theis999: It triggers on right-clicking the move up/down buttons. It is very very slow, which might cause some confusion.

src/cMain.cpp
theis999: Fair. Which do you prefer spin_<name> or <name>_spin => spin_x vs x_spin

MortenTobiasNielsen: spin_x

src/cMain.cpp
theis999: Resolved in #93

src/cMain.cpp
theis999: Done

src/cMain.cpp
theis999: I must admit that i liked the scrolling but it wasn't accurate enough so incidentally you get your way. Solved with #93

src/cMain.cpp
theis999: Solved with #93 which is still not end of implementing search

theis999:
There are handlers for resizer events, but i don't want to dig into that headache.

I have adjusted the column widths to the best of my abilities.
I have added search implementation to separate PR (this one is getting long)
I have added tooltips to indicate move 5. 

Some multimove of task is required and 5 seems like a good number in my opinion. I have moved the warning and update-building-grid outside the internal move, which also made move 5 faster.

PR #83, opened 2022-09-28
Moved Save to a separate class

MortenTobiasNielsen:


src/SaveTas.cpp
```diff
@@ -1,7 +1,130 @@
 #pragma once
 
 #include "SaveTas.h"
+#include "utils.h"
```
theis999, line 4: Couldn't you move utils.h to savetas.h?

theis999 (approved):
It looks like good. A few minor comments.

PR #84, opened 2022-09-30
Feature/move open

MortenTobiasNielsen:


src/OpenTas.cpp
```diff
@@ -2,6 +2,278 @@
...
+	return true;
+}
+
+bool OpenTas::extract_script_location(std::ifstream& file) {
+	if (!update_segment(file) || seglist[0] != code_file_indicator && seglist[0] != "Task folder location:") {
+		return false;
+	}
+	
+	if (!update_segment(file) || seglist.size() != 1) {
+		return false;
```
theis999, line 191: It looks like you evaluate !update_segment twice. Maybe these two if statements be merged

src/OpenTas.cpp
```diff
@@ -2,6 +2,278 @@
...
+	return true;
+}
+
+bool OpenTas::extract_auto_close(std::ifstream& file) {
+	if (!update_segment(file) || seglist[0] != auto_close_indicator) {
+		return false;
+	}
+
+	if (!update_segment(file) || seglist.size() != 2 || seglist[0] != auto_close_generate_script_text) {
+		return false;
```
theis999, line 204: It looks like you evaluate !update_segment twice. Maybe these two if statements be merged

src/cMain.h
```diff
@@ -154,10 +154,7 @@ class cMain : public GUI_Base {
 	Shortcuts_Menu* shortcuts = nullptr;
 	dialog_progress_bar_base* dialog_progress_bar = nullptr;
 
-	std::string goal_steelaxe_text = "Steel Axe";
-	std::string goal_GOTLAP_text = "Getting On Track Like A Pro";
-	std::string goal_any_percent_text = "Any %";
-	std::string goal_debug_text = "Debug";
+
```
theis999, line 157: This looks like excessive empty line

theis999 (changes requested):
Magic numbers (10, 11 etc) should be put in a define structure or similar.

src/OpenTas.cpp
```diff
@@ -2,6 +2,278 @@
...
+	return true;
+}
+
+bool OpenTas::extract_goal(std::ifstream& file) {
+	if (!update_segment(file) || seglist[0] != goal_indicator) {
+		return false;
+	}
+
+	if (!update_segment(file) || seglist.size() != 1) {
+		return false;
```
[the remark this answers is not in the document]
MortenTobiasNielsen: The first one evaluates that the segment has the correct header and the next one evaluates that the steps underneath the heading is correctly formatted and if that is the case goal is set to that segment.

src/OpenTas.cpp
MortenTobiasNielsen: Every time update_segment is called it goes to a new line, so the two calls is needed to evaluated two different lines

src/OpenTas.cpp
MortenTobiasNielsen: Every time update_segment is called it goes to a new line, so the two calls is needed to evaluated two different lines

src/OpenTas.cpp
```diff
@@ -2,6 +2,278 @@
...
+
+			return false;
+		}
+
+		if (template_name == "") {
+			template_name = seglist[0];
+			template_list = {};
+		}
+		else if (template_name != seglist[0]) {
+			return_data.template_map.insert(std::pair<std::string, std::vector<std::string>>(template_name, template_list));
```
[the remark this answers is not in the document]
MortenTobiasNielsen: How would you refactor it? template_name and template_list should only be reset if the template_name is "" or if the template_name differs from seglist[0] and the template map should only be reset if template_name differs from seglist[0].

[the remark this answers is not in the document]
theis999: Looking at it again i think you have done it correctly.

src/OpenTas.cpp
theis999: I see. I guess that would be slightly more obvious if the function was called next_line or next_segment.

theis999 (approved):

src/Functions.h
```diff
@@ -12,4 +14,6 @@ void string_capitalized(std::string &string);
 
 void find_coordinates(std::string& x_cord, std::string& y_cord, std::string& direction, std::string& building_size);
 
-bool is_number(const std::string& s);
\ No newline at end of file
+bool is_number(const std::string& s);
```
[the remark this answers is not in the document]
MortenTobiasNielsen: I don't know, but that might be. 😊

PR #89, opened 2022-11-10
Fix to splitter fast replace

theis999:
It seems like player.surface.create_entity increases step by 1 usually, but with a splitter it increase by 2.

### DIFF
--- a/Lua Files/control.lua
+++ b/Lua Files/control.lua
@@ -319,8 +319,10 @@ local function build()
 
 		elseif player.can_place_entity{name = item, position = target_position, direction = direction} then
 			if player.surface.can_fast_replace{name = item, position = target_position, direction = direction, force = "player"} then
+				local save_step = step
 				if player.surface.create_entity{name = item, position = target_position, direction = direction, force="player", fast_replace=true, player=player, raise_built = true} then
-					change_step(-1)
+					save_step = save_step - step
+					change_step(save_step)
 					player.remove_item({name = item, count = 1})
 					return true
 				end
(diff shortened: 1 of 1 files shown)


MortenTobiasNielsen:
After some testing it turns out that fast replace triggers a mining event (which increases the step count by 1) per entity mined, so if a splitter replaces two times transport belt it will trigger two mining events and increase step count by 2. I hadn't tested that case when I developed it. 

I would have liked to solve the problem at its root (this mining event), but it doesn't seem like the event has an indicator for whether it is triggered by fast replace, so I don't really know how to do that. :( Would you add your solution to the rail section as well? I don't think it is possible to replace multiple pieces of rail at the same time, but just in case.

theis999:
Credit should go zaspar.

I had been wondering why mining entities increased the step for a while. It gives some odd behaviors when the TAS script isn't running and you have the task list from my mod open.

I added the fix for rails too, but i do not know how to test it. I have never worked with rails in the tas generator.

There might be a way using [LuaPlayerMinedEntityEventFilter](https://lua-api.factorio.com/latest/Concepts.html#LuaPlayerMinedEntityEventFilter) but it is not obvious. 

Due note that fast replacing multiple belts with a set of underground belts doesn't work. Which likely means that pipe-to-ground also doesn't work.

MortenTobiasNielsen:
Yeah, that does look a little odd when that happens. :)

I don't really know how to test that either, but I am pretty sure that it wouldn't be used in a TAS. Actually using trains seems pretty inefficient for a speed run.

It could be a thing, but yeah it is not obvious how it should be used.

I will test it out and see what is causing that behavior.

MortenTobiasNielsen:
I guess we could add a "Only change_step when run is true" check for the mining event, so it doesn't trigger your mod.

MortenTobiasNielsen:
You are correct underground belt and pipe-to-ground doesn't remove the belt/pipe between, I unfortunately have no clue how to solve that problem. To me it goes into the "it is used very rarely" pile. I won't be spending time trying to fix that unless it become really annoying to someone.

theis999:
> I guess we could add a "Only change_step when run is true" check for the mining event, so it doesn't trigger your mod.

That does appear to work.

PR #91, opened 2022-11-11
mining-event-only-on-mining-task

theis999:
Mining removed changed step when fast replacing
Added condition so change step on-mining only happens when the tas is running and the current step is a mining task.


MortenTobiasNielsen:
Very nice - I will merge it in and test it together with my own changes.

theis999:
Thank you. It was way easier when I understood the problem :)

PR #92, opened 2022-11-14
Refactor type panel

theis999:
This is almost a pure refactor building on the GUI overhaul branch

### Refactor
Moved event handlers for radio buttons
Moved event handlers for menu-shortcuts related to radio buttons
Moved extract_task
Moved setup_parameters

### Changes
Added a TypePanel class
Converted some static strings that were used to manipulate the GUI into named variables
Cleaned extract_task


theis999:
It is worth to not 2 things: 

C++ allows the implementation of any declared function to be place anywhere. Which is why TypePanel.cpp can implement cMain's functions.

The line `friend TypePanel;` added in #0810c7e allows TypePanel to access variables from cMain eventhough they are protected.

PR #93, opened 2022-11-18
Search expansion

theis999:
Moved search algorithm to seperate file.

Extended search to handle semicolons as separate search terms AND'ed together and colons to specify columns. 
If no colon then it tries to defaults to search for either text or numbers.

Search:
`a`: search for a in the text columns. `Idle`
`x`: search for x in the number columns. `-3.1`
`A:a` search for a in column A. `Task:Idle`
`A:a;b` search for row with a in column A and b in any text columns. `Task:Build;North`

Notes:
The column is not case sensitive but the field text is.
The scrolling has been changed to a jump.


theis999:
Rebased on master branch.

PR #94, opened 2022-11-18
Walking improvement

theis999:
Added buffer for intermediate walking tasks
Improved rounding function in lua for determining position and destination
Added a way to handle walkinging task tickless, if we are already at the destination.


theis999:
I seem to have made a bug where the tickless walk, skips a walk command if there are 2 walk commands in a row.

Issue #90, opened 2022-11-11
Fast replacing underground belt and pipe to ground


theis999:
The game allows the user remove belts/pipes in between two undergrounds while fast-replacing. The TAS currently doesn't.

theis999:
#95 closes

PR #95, opened 2022-11-20
Custom fast replace underground

theis999:
### Custom handling of fast replacing underground-belt and pipe-to-ground with belts/pipes in between
As mentioned in #90 fast replacing belts/pipes wasn't working correctly
It turns out it is a missing feature [create_entity](https://lua-api.factorio.com/latest/LuaSurface.html#LuaSurface.create_entity) where the field fast_replace promises `If true, building will attempt to simulate fast-replace building.` which doesn't hold true for creating a pair of undergrounds with belts or pipes inbetween. 

This PR adds custom code to fix that shortcoming. I will also create a bug report for the factorio team, so hopefully we will have a better solution in the future.

Note: since the fix to splitters, we no longer care about if a build action is a fast-replace or not.

### DIFF
--- a/Lua Files/control.lua
+++ b/Lua Files/control.lua
@@ -277,6 +277,65 @@ local function tile_is_in_reach()
 	return dis <= 10.25 -- It seems like 10.25 aligns best with the current walking algorithm
 end
 
+---comment Places an entity, possibly fast replacing. Then handless 
+---@return boolean true if an entity is created.
+local function create_entity_replace()
+
+	local fast_replace_type_lookup = {
+		["underground-belt"] = {"transport-belt", "fast-transport-belt", "express-transport-belt"},
+		["fast-underground-belt"] = {"transport-belt", "fast-transport-belt", "express-transport-belt"},
+		["express-underground-belt"] = {"transport-belt", "fast-transport-belt", "express-transport-belt"},
+		["pipe-to-ground"] = {"pipe"}
+		}
+
+	local created_entity = player.surface.create_entity{name = item, position = target_position, direction = direction, force="player", fast_replace=true, player=player, raise_built = true}
+	if created_entity and fast_replace_type_lookup[created_entity.name] ~= nil and created_entity.neighbours  then --connected entities eg underground belt https://lua-api.factorio.com/latest/LuaEntity.html#LuaEntity.neighbours
+		local replace_type = fast_replace_type_lookup[created_entity.name]
+
+		local neighbour_position = nil
+		if (#created_entity.neighbours == 0) then 
+			neighbour_position = created_entity.neighbours.position
+		else
+			for _,neighbour in pairs(created_entity.neighbours[1]) do
+				if (created_entity.name == neighbour.name) then
+					neighbour_position = neighbour.position
+				end
+			end
+		end
+		if (not neighbour_position) then
+			player.remove_item({name = item, count = 1})
+			return true
+		end
+
+		local entities_between = player.surface.find_entities_filtered{area = {created_entity.position, neighbour_position}, name = replace_type}
+		local entities_between_length = math.abs(created_entity.position.x - neighbour_position.x + created_entity.position.y - neighbour_position.y) - 1
+		local can_replace_all = entities_between_length == #entities_between
+
+		--chech that all entities betweeen are in the same direction
+		if can_replace_all and not created_entity.name == "pipe-to-ground" then --ignore direction for pipes
+			for __, e in pairs(entities_between) do
+				if e.direction ~= created_entity.direction then
+					can_replace_all = false
+					break
+				end
+			end
+		end
+		--mine all entities inbetween
+		if can_replace_all then
+			for __, e in pairs(entities_between) do
+				e.mine{player.character.get_main_inventory(), true, true, false}
+			end
+		end
+		--spend the item placed
+		player.remove_item({name = item, count = 1})
+		return true
+	end
+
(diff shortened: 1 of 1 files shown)


Lua Files/control.lua
```diff
@@ -277,6 +277,65 @@ local function tile_is_in_reach()
...
+				e.mine{player.character.get_main_inventory(), true, true, false}
+			end
+		end
+		--spend the item placed
+		player.remove_item({name = item, count = 1})
+		return true
+	end
+
+	--no special fast replace handling
+	player.remove_item({name = item, count = 1})
```
MortenTobiasNielsen, line 335: I don't think we can get into a situation where we won't create the entity, but I think we should check that the entity was actually created before we remove it from the users inventory.

Lua Files/control.lua
```diff
@@ -277,6 +277,65 @@ local function tile_is_in_reach()
...
+		["express-underground-belt"] = {"transport-belt", "fast-transport-belt", "express-transport-belt"},
+		["pipe-to-ground"] = {"pipe"}
+		}
+
+	local created_entity = player.surface.create_entity{name = item, position = target_position, direction = direction, force="player", fast_replace=true, player=player, raise_built = true}
+	if created_entity and fast_replace_type_lookup[created_entity.name] ~= nil and created_entity.neighbours  then --connected entities eg underground belt https://lua-api.factorio.com/latest/LuaEntity.html#LuaEntity.neighbours
+		local replace_type = fast_replace_type_lookup[created_entity.name]
+
+		local neighbour_position = nil
+		if (#created_entity.neighbours == 0) then
```
MortenTobiasNielsen, line 296: What does the # do?

Lua Files/control.lua
```diff
@@ -277,6 +277,65 @@ local function tile_is_in_reach()
...
+		["pipe-to-ground"] = {"pipe"}
+		}
+
+	local created_entity = player.surface.create_entity{name = item, position = target_position, direction = direction, force="player", fast_replace=true, player=player, raise_built = true}
+	if created_entity and fast_replace_type_lookup[created_entity.name] ~= nil and created_entity.neighbours  then --connected entities eg underground belt https://lua-api.factorio.com/latest/LuaEntity.html#LuaEntity.neighbours
+		local replace_type = fast_replace_type_lookup[created_entity.name]
+
+		local neighbour_position = nil
+		if (#created_entity.neighbours == 0) then 
+			neighbour_position = created_entity.neighbours.position
```
MortenTobiasNielsen, line 297: Can we get the position if there is no neighbours?

Lua Files/control.lua
```diff
@@ -277,6 +277,65 @@ local function tile_is_in_reach()
...
+		else
+			for _,neighbour in pairs(created_entity.neighbours[1]) do
+				if (created_entity.name == neighbour.name) then
+					neighbour_position = neighbour.position
```
MortenTobiasNielsen, line 301: Are neighbour_position an array of neighbour positions or is it finding the neighbour furthest away? What would happen if I build underground belt further away than what would make sense to do e.g. 4 tiles for yellow belt (I think). :)

MortenTobiasNielsen (commented):
I have a few questions, mostly for my understanding. 

I am getting some kind of naming error, which I am pretty sure is handled in the master branch, so could you merge master into the branch?

Lua Files/control.lua
theis999: Oh right, i think we can. It should in theory fail if the location is blocked by water or another entity.

Lua Files/control.lua
theis999: It is like size() or length()

Lua Files/control.lua
MortenTobiasNielsen: Won't "can_place_entity" fail, so we don't get here, if that is the case?

Lua Files/control.lua
MortenTobiasNielsen: I see - not really obvious, but I guess it can be nice with a short hand. :)

Lua Files/control.lua
theis999: It is sort of stupid, but neighbours is an entity when it is an underground-belt and an array of entities when it is a pipe-to-ground. However we know from a statement above this, that neighbours ~= nil. So the if statement actually test if it is an array or not.

Lua Files/control.lua
theis999: neighbour_position is a local variable (a single point{x,y})
created_entity.neighbours is an array with an with the first element being an array of connected entities. For pipes this means other entities with fluid boxes like pipes or boilers. If it is further away than it can connect then it doesn't show up in this array.

theis999:
I didn't see any naming error after rebasing so i guess it is fine now.

Lua Files/control.lua
theis999: You are correct, anyway i added the check. It likely won't matter to performance compared to the rest of the function.

theis999:
It turns out that i should have used player.mine_entity instead of entity.mine

MortenTobiasNielsen:
I just realized that I never corrected the error on the main branch, but just on the branch I am currently developing on. I will just make the change locally so I can test it. 

When this is merged in I will get my own branch aligned with master, which is going to be absolute hell - but that is what happens when a branch gets too old.

theis999:
Rebased on master.
Added check to pipe-to-ground if there is pipe split between, it won't replace the pipes between.

PR #104, opened 2022-11-25
Fixed pick up

### LINKED ISSUE #58: Pickup is missing walking
As is, the pickup task, runs for 1 tick while you are forced to stand still.

In regular factorio, 
- it is something you can do while walking/mining/shooting etc. 
- it is something you do for awhile

MortenTobiasNielsen:
I couldn't see a way to make it backwards compatible due to how wrong I had implemented it. I hope that a run can be re-aligned with a combination of walking, picking and idle tasks. 

closes #58

### DIFF
--- a/Lua Files/control.lua
+++ b/Lua Files/control.lua
@@ -8,6 +8,7 @@ local run = true
 local step = 1
 local step_reached = 0
 local idle = 0
+local pick = 0
 local mining = 0
 
 local player
@@ -827,7 +828,7 @@ local function doStep(steps)
 		target_position = steps[3]
 		amount = steps[4]
 		slot = steps[5]
-		
+
 		return limit()
 
 	elseif steps[2] == "priority" then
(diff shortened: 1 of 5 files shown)


Lua Files/control.lua
```diff
@@ -856,13 +857,14 @@ local function doStep(steps)
 		return drop()
 
 	elseif steps[2] == "pick" then
+		pick = steps[3] - 1
```
theis999, line 860: Maybe `pick = pick + steps[3] - 1` in case of two of pick up task in sequence.

Lua Files/control.lua
```diff
@@ -8,6 +8,7 @@ local run = true
 local step = 1
 local step_reached = 0
 local idle = 0
+local pick = 0
```
theis999, line 11: I would rather have a more qualified named, like pickup_ticks.

theis999 (approved):
Looks like a solid conversion to a much more correct implementation 👍

Lua Files/control.lua
MortenTobiasNielsen: Agreed.

Lua Files/control.lua
MortenTobiasNielsen: Agreed.

PR #111, opened 2022-12-08
Refactor buildingNameToIndex

theis999:
Renamed (building)Index to type
Renamed (building)OrientationIndex to orientation
Renamed BuildingNameToIndex to BuildingNameToType

Changed BuildingNameToType to not be const, allowing the syntax BuildingNameToType[name]. Which i find looks nicer.

Added vector BuildingNames,
Changed FindBuildingName to use BuildingNames for instant lookup


src/Building.h
```diff
@@ -6,14 +6,14 @@ struct Building
 {
 	double X;
 	double Y;
-	int Index;
-	Orientation OrientationIndex;
+	int type;
```
MortenTobiasNielsen, line 9: This is a little confusing. This is not a type it is an integer.

src/BuildingNameToIndex.cpp
```diff
@@ -2,13 +2,6 @@
...
-	for (auto it = BuildingNameToIndex.begin(); it != BuildingNameToIndex.end(); ++it)
-	{
-		if (it->second == index)
-		{
-			return it->first;
-		}
-	}
-
-	return "N/A";
+	if (index < 1 || index > BuildingNames.size()) return "N/A";
```
MortenTobiasNielsen, line 5: Please use the same code style as the rest of the code base. :)

src/BuildingNameToIndex.h
```diff
@@ -21,7 +29,108 @@ enum Assemblers
 	Centrifuge,
 };
 
-static const std::map<string, int> BuildingNameToIndex =
+static inline vector<string> BuildingNames = {
+	"NONE", // for alignment
```
MortenTobiasNielsen, line 33: How is this supposed to be used?

MortenTobiasNielsen (commented):
Just have a few comments. :)

src/Building.h
theis999: Index was an integer that referenced which type of building it was, not which number of building it was. 
Effectively it is an enumeration, so naming the variable type is appropriate.

src/BuildingNameToIndex.h
theis999: We need this to be true: `BuildingNames[index] = name -> BuildingNameToType[name] = index`
But the first element in BuildingNameToType has the value index=1, so the array BuildingNames needed to be padded 1 in the start.

PR #140, opened 2022-12-28
Cancel crafting

theis999:
Added cancel crafting command to the GUI
Added cancel crafting step handler to lua

The GUI allows you to cancel crafting X amount of item Y
The Lua code scans the crafting queue for item Y

1.  If the scan finds item Y
- the amount is -1 then it cancels all
- the amount is less than or equal to crafting count, it cancels those
- the amount is more than what is being crafted, it throws a warning
2. If the scan doesn't find item Y, then it throws a warning 

This was a rather fast implementation, some adjustments might needed. I can see that the name "Cancel Crafting" is a bit long on the GUI. It wasn't clear from the issue how exactly it should be handled, when scanning for the desired craft to cancel.

### DIFF
--- a/Lua Files/control.lua
+++ b/Lua Files/control.lua
@@ -312,6 +312,28 @@ local function craft()
 	end
 end
 
+-- Cancels a handcraft a recipe
+local function cancel_crafting()
+	local queue = player.crafting_queue
+
+	for i = 1, #queue do
+		if queue[i].recipe == item then
+			if count == -1 then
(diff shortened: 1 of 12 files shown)


MortenTobiasNielsen:
Lets call it Cancel. Until we get some other type of cancel step I think it is fine to do that.

theis999:
Done

MortenTobiasNielsen (approved):

PR #141, opened 2022-12-29
Convert chat commands to console commands

theis999:
Simple change. The user needs to use "/release" instead of "release", but also allows them to use tap to auto-complete.

Added skip x, that allows the user to go back or forth x steps. It still defaults to x = 1.

### DIFF
--- a/Lua Files/control.lua
+++ b/Lua Files/control.lua
@@ -1481,18 +1481,22 @@ end)
 
 script.on_load(migrate_global)
 
-script.on_event(defines.events.on_console_chat, function(event)
-	if event.message then
-		if event.message == "release" then
-			run = false
-		end
+local function release()
+	run = false
+end
 
-		if event.message == "resume" then
-			run = true
-		end
+local function resume()
+	run = true
+end
 
-		if event.message == "skip" then
-			change_step(1)
-		end
+local function skip(data)
+	if data and data.parameter and tonumber( data.parameter ) then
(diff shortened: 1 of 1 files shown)


MortenTobiasNielsen:
There unfortunately seem to be some issues.

1. I am not able to use a saved game
2. Theis TAS Helper crashes if I try to skip a step when non is available

theis999:
1. I don't understand what you mean? i haven't changed anything related to saving
2. Fair, i don't expect that has changed from what was before. But i will handle it the next time a look at my mod.

MortenTobiasNielsen:
1. The game crashes when I try to load a saved game (in factorio). It had something to do with the mod generated by the TAS tool. I really should have taken a screen shot. I will have a look at it and provide some more info later today or tomorrow.
2. Agreed, I just didn't test that edge case and to be fair it is not totally expected and I don't think users will actually do that, so it is mostly a nice to have.

theis999:
1. I haven't seen any issues with resuming from saved game
2. I mean, this change does make it more likely that the user overflows the steplist. Since `/skip -1000` and `/skip 1000` are now possible.

MortenTobiasNielsen:
1. I can't recreate the problem. I guess it most have had something to do with jumping between a lot of branches or something like that. 
2. True. The user might think that they need to move more steps than what is needed or mistype.

MortenTobiasNielsen (approved):

Issue #123, opened 2022-12-15
Wait until the specified amount is available on "Take"


MortenTobiasNielsen:
I often find that I am using an unreasonable amount of time trying to time the last few ticks of a furnace or assembling machine. It would be so much easier if it just waited until the item was available, if the character is running - like it is done for buildings. It would also solve some of those "1 tick of" misalignment problems which occurs some times.

theis999:
I agree. Though it would cost you ticks, so we should be careful.

Some details:
This only applies for take `x > 0`, since you need an amount.
With take All, then it just requires 1 item to be taken.
Since it is bound to cost ticks, i think it should be left up to the user.
So we need another field, `wait_for_items` or similar.

MortenTobiasNielsen:
The only time this would take effect is if the developer has specified that they want to take all or a very specific amount. In most cases the run will break if it isn't available, so I rather have it wait 1 tick due to a misalignment and still work than it breaking. In almost all cases the amount will be available, so it won't do this. I don't really want to make an extra field to specify this it should just work.

theis999:
So this applies to Take, Put, and Drop. I still don't think Drop matters much so we will ignore it for now.

For Take x > 0 and Put x > 0, you want to wait till you have x.
For Take ALL and Put ALL, you want to wait till you have at least 1.

What do you want to do in a multibuilding step. Would it be fine if Take x=3 for 4 buildings, that it took 3 for the first two buildings and only took 1 for the last two buildings?

theis999:
Understood. I expect you also want a warning/debug statement.

I expect this to be rather simple change.

PR #144, opened 2023-01-02
Wait till amount is availble

theis999:
#123 Wait until items are availble

Changed Put, Take and Craft to wait for at least one item when using X = ALL
Changed Put, Take and Craft to wait for X if X is != ALL

Changed drop to wait for an item
Changed drop to remove the dropped item from the inventory

I haven't tested drop, as it was broken in c++ part. Craft, take and put have been verified.

### DIFF
--- a/Lua Files/control.lua
+++ b/Lua Files/control.lua
@@ -216,10 +216,10 @@ local function put()
 	if removalable_items == 0 or insertable_items == 0 then
 		if removalable_items == 0 then
 			warning(string.format("Step: %s, Action: %s, Step: %d - Put: % is not available in your inventory", task[1], task[2], step, item))
-	end
+		end
 		if insertable_items == 0 then
 			warning(string.format("Step: %s, Action: %s, Step: %d - Put: % can't be put into target inventory", task[1], task[2], step, item))
-	end
+		end
 		return false
 	elseif amount < removalable_items or amount < insertable_items then
 		debug(string.format("Step: %s, Action: %s, Step: %d - Put: not enough % can be transferred", task[1], task[2], step, item))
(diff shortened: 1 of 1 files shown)


theis999:
I changed it to warnings, which shouldn't spam quite as much

MortenTobiasNielsen:
Sorry, but I don't want warnings either. Unless the character stops running there is no problem in waiting for some ticks before executing a step - nothing is lost by that, given that the ticks would just be "wasted" afterwards as the character is running without any steps to execute.

As a developer I would though like to get a warning telling me how many ticks was wasted waiting to executed a step, if the character stops with steps still pending for that run sequence.

theis999:
That seems substantially different from the original description.

1. We currently don't have the ability to choose between **debug** and **warnings** messages. And a message of waiting for a number items before proceeding, seems perfect for debug _level_ msg.
2. We don't currently have the ability to create **delayed messages** or measure the tick between **starting** a task and **ending** it. It does sound useful for a number of other things.

I believe it is important to give the developer a heads up, when the TAS does something different than what they might expect. 
- I can see the case, where waiting 3 ticks for an item might not warrant a warning, a verbose msg would fit here. 
- Waiting 120 ticks or _forever_, should give a warning. 
- If the character is walking, it might walk past the item source, which could be problematic and should give a debug msg.

I think it is out of scope for this PR to solve 1&2, we could wait for a future where they are solved or work with what we have now. In the latter case, we have to either print a warning message that could be annoying or not print a message which could be confusing.

MortenTobiasNielsen:
I am pretty sure that I didn't say that I want the user to be spammed with messages when the feature is used. :) It is the same behavior as we have with building something. If the buildings isn't available to build the character will keep on running until it either has the building available and it is build or the characters stops which then triggers a warning - because that most likely wasn't the intend to have the character just stand around waiting. 

It isn't possible for the character to run past something, given that the tool will ensure that the character won't be out of reach to perform a step. I don't think that waiting for 120 ticks should give a warning as long as the character is running - until the character stands still I expect that the developer have chosen to do it - and it isn't a waste of ticks (unless the character is standing still).

MortenTobiasNielsen (approved):

PR #146, opened 2023-01-03
fix to open tas when generator file is missing

theis999:
**Fix** to open tas file throwing error when generator file is missing
**Fix** to save tas file skipping when the generator file is missing

Removed wxWindow param from open tas, as it wasn't used
Added documentation comment to open tas and save tas

### DIFF
--- a/src/Factorio-TAS-Generator.vcxproj
+++ b/src/Factorio-TAS-Generator.vcxproj
@@ -196,6 +196,7 @@
     <ClInclude Include="DialogProgressBar.h" />
     <ClInclude Include="GridEntry.h" />
     <ClInclude Include="StepParameters.h" />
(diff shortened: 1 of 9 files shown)


theis999:
@MortenTobiasNielsen I have added some substantiel updates to this branch.

I have consolidated the use of JSON into a single file called `Settings.h`
I have added setting struct to handle the data parsing to other parts of the project
I have added UTF-8 conversion (blatantly stolen from https://www.codeguru.com/cplusplus/the-basics-of-utf-8/)
I have applied the conversion (only to the last-tas field)

MortenTobiasNielsen (approved):
Seems to work and I think it is nicer that it is mostly contained to its own file - good work.

PR #148, opened 2023-01-07
Aligned Put, Take and Craft with the "keep running" logic in Build

MortenTobiasNielsen:
changed warning to provide extra info.


theis999 (approved):
Looks reasonable 👍

PR #157, opened 2023-02-02
Feature/pick up backwards compatibility

### LINKED ISSUE #155: Pickup (f) doesn't work in compatibility mode
When in "old" mode pickup_ticks is never set, so the tas doesn't know it is in pickup state. However it also doesn't progress the step counter forward, so the character just stands there.

MortenTobiasNielsen:
Closes #155 

 It is by no means pretty, but it can be used for backwards compatibility - Zaspar tried it on his run and it seems to work.

### DIFF
--- a/Lua Files/control.lua
+++ b/Lua Files/control.lua
@@ -1059,6 +1059,10 @@ local function doStep(current_step)
 		drop_item = current_step[4]
 		return drop()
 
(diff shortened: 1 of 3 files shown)


theis999 (approved):
If it works then it is fine 👍

PR #168, opened 2023-02-11
Fix put module

theis999:
Reversed the order, of insert and remove items to make sure we can add all of the items we want before removing them.

### DIFF
--- a/Lua Files/control.lua
+++ b/Lua Files/control.lua
@@ -237,12 +237,16 @@ local function put()
 		return false
 	end
 
+	amount=target_inventory.insert{
+		name=item,
+		count=amount,
+	}
+
+	if amount == 0 then return false end
+
 	amount = player.remove_item{
 		name=item,
-		count=target_inventory.insert{
-			name=item,
-			count=amount,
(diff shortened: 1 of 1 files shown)


MortenTobiasNielsen:
Seems like we need some extra safe guards:
image

It happened while putting (32 iron plates into an assembling machine). Not sure why the code change would cause a negative number.

theis999:
I couldn't easily force the issue. My concern is at this:

image
If either removalable_items or insertable_items returns a negative value. Then math.min also returns a negative value.

MortenTobiasNielsen:
Surely, but in what scenario would the character carry a negative amount or the buildings inserting into only be able to receive a negative amount?

theis999:
I honestly don't see how any of the code could cause the crash you experienced. Unless it isn't a negativ value but just a zero?

theis999:
image
Step 8 - crashes the game like you reported

MortenTobiasNielsen (approved):
Nice to get the warning, so it is clear that the step isn't configured correctly.

Issue #165, opened 2023-02-11
Paint commands that adds walk steps


theis999:
I can see that Zaspar has trouble figuring out when a Walk step gets added by the generator.

It would be nice to indicate when a command adds a walk step, so it is visible in the generator without running the game.
That way it would also be easier to pin point problematic tasks.

One way could be to color a part of the grid line, on generate script.

MortenTobiasNielsen:
I agree that this is a problem. I usually solve it by opening the script and see if there are walk steps with impossible coordinates, which clearly isn't a good solution. 

I guess it would be possible to run the distance calculation on each step added and color the text red if it is not within reach, I am though a little in doubt users will understand why it is happening.

theis999:
Yeah I would like deal with it the same way, but we are both programmers with integral knowledge of the system 😎

There are many ways to deal with it to a more general user. 
I like the idea of painting: 
- It would allow us to target the problematic tasks visually
- It doesn't block the user (like a pop-up)
- It is easy to see when scrolling a huge list of steps

But your concern is valid

MortenTobiasNielsen:
Lets go with the painting solution and hope users doesn't get too confused. :)

PR #170, opened 2023-02-12
Paint substep walk

theis999:
#165  Paint steps in steplist when extra substep walks are added.

It is perhaps not the pretties:
image

Also added a translation map to technologies.

### DIFF
--- a/src/GenerateScript.cpp
+++ b/src/GenerateScript.cpp
@@ -59,7 +59,7 @@ void GenerateScript::AddInfoFile(string& folder_location)
 void GenerateScript::PaintWalk(string step)
 {
 	int row = std::stoi(step) - 1;
-	grid_steps->SetCellBackgroundColour(row, 9, *wxRED);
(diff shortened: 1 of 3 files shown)


MortenTobiasNielsen:
For some reason I haven't noticed this pull request - sorry. I think it would be nice if it was painted red and a comment like "Walk step added" was there - which should be "caught" in lua, so it isn't printed out.

theis999:
I can paint it red, i don't know if agree that it should be Error color.
I can't add a comment, as that would either override a comment that is already there or adds a comment that shouldn't be there. Of course _i can_, i just don't think it is a good idea.

MortenTobiasNielsen:
Agreed - wasn't the best ideas. :D

MortenTobiasNielsen (approved):
I am still a little worried that the user won't really know why the comment column is colored, but I on the other hand have no better alternative, so lets try it out. :)

Nice to get the tech into a translation table.

PR #174, opened 2023-03-05
Fixed drop command

theis999:
### Fixed drop command
Removed **multibuild** from drop command
Removed **row_drop**
Removed **building checks** from drop steps
Changed **drop step** to always use **"Item on ground"** for reach calculation

Changed **tas.drop** to always use a custom reach calculation.

### DIFF
--- a/Lua Files/control.lua
+++ b/Lua Files/control.lua
@@ -933,7 +933,7 @@ end
 
 -- Drop items on the ground (like pressing the 'z' key)
 local function drop()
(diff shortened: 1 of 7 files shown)


theis999:
Do note that:
-  drop does not allow dropping items into buildings, but the user can just use put for that
-  drop does allow dropping two "item-on-ground" at the same locations, as the [spill](https://lua-api.factorio.com/1.1.74/LuaSurface.html#LuaSurface.create_entity) parameter of create_entity does not overflow items into nearby areas.

MortenTobiasNielsen (approved):
Seems to work nicely. Really nice to get that done - thanks. :)

PR #177, opened 2023-03-07
Refactor: Parameter choice

theis999:
This is a refactor and any functional changes are unintended.

The main objective was to
1. getting a more simple way to specify parameters types
2. going away from using unnamed indexes to specify parameter enabling

It is heavy on bit-wise operations [  &, |, <<  ], the most confusing of them is 1<<N
```
1 << 0 == 0001
1 << 1 == 0010
1 << 2 == 0100
```
I have also used chained variable constructors, which helps IntelliSense adding a comment to each of the variables:
```
        //game interactions
	const int Pause = comment,
	stop = comment,
	save = comment,
	game_speed = amount | comment;
```
image

My favorite part happens in the internal part, where common patterns are specified.

### DIFF
--- a/src/Factorio-TAS-Generator.vcxproj
+++ b/src/Factorio-TAS-Generator.vcxproj
@@ -186,6 +186,7 @@
     <ClInclude Include="GenerateScript.h" />
     <ClInclude Include="GUI_Base.h" />
     <ClInclude Include="nlohmann\json.hpp" />
+    <ClInclude Include="ParameterChoices.h" />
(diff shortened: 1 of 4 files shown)


MortenTobiasNielsen (approved):
Nice work this is very clever. It is though also very overkill for the use case. It seems like bitwise operator "magic" should only be used when doing compute heavy things. It makes the code relatively hard to read without any tangible benefit (in this case), but not so much that it is too problematic, so I will allow it. :)

PR #179, opened 2023-03-07
Removed CheckTakePut(std::string& item) as it wasn't defined

theis999:


MortenTobiasNielsen (approved):
Nice that you found this, must have left it by mistake when I did the large refactor. :)

PR #180, opened 2023-03-07
Refactor of general gui elements

theis999:
Changed the name of a number of gui items, to make it easier to navigate in the wxFormBuilder
Added tooltips to all Commands/step_types, to make it easier for new users

PR #181, opened 2023-03-07
Step painter

theis999:
Demo showing the effect of painting rows in the step list.

A colour picker was added to the auto put panel, mainly due to the ample amount of space there.
An event handler for updating the colour value of the colour picker is implemented.
An event handler for updating the background colour of selected rows in the step list.

To complete, the colour of rows needs to be stored in the save file.
Due note that this is built on #180

### DIFF
--- a/src/OpenTas.cpp
+++ b/src/OpenTas.cpp
@@ -117,7 +117,7 @@ Category OpenTas::extract_steps(std::ifstream& file, DialogProgressBar* dialog_p
 	while (update_segment(file))
 	{
 		if (segments.size() != step_segment_size && 
-			segments.size() != step_segment_size_without_comment &&
+			segments.size() != step_segment_size_without_colour &&
 			segments.size() != step_segment_size_without_comment_and_colour)
 		{
 			if (segments[0] == save_templates_indicator)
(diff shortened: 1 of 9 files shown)


theis999:
image

MortenTobiasNielsen:
I guess it cannot hurt to add this, so the user has the possibility to use it. :)

theis999:
I have added **save/load** functionality

In the switch case of step parameters i also added [[[likely]]](https://en.cppreference.com/w/cpp/language/attributes/likely) which is something i stumbled upon that should optimize the compilation for those cases.

src/OpenTas.cpp
```diff
@@ -163,6 +165,7 @@ Category OpenTas::extract_steps(std::ifstream& file, DialogProgressBar* dialog_p
 		step.Orientation = Capitalize(segments[5]);
 		step.Direction = Capitalize(segments[6]);
 		step.Comment = comment;
+		step.Colour = segments[10];
```
MortenTobiasNielsen, line 168: Colour won't be in the current save files, so this will crash the application.

MortenTobiasNielsen (changes requested):
You will have to handle colour in the same way as I have handled comment, so it is only extracted if it is there.

theis999:
Alright, i changed some stuff and tested on file without colours, and one without colours and comments.

MortenTobiasNielsen (approved):

PR #182, opened 2023-03-07
Improvement to change Command type

### LINKED ISSUE #178: Changing from PUT to TAKE unsets Item

theis999:
**Added** struct that holds **pointers** to the current version of underlaying gui elements.
**Added** methods to handle **update** of gui elements
**Changed** change task to use the new methods

This fixes #178 by only **updating** the item list, if it is **different**. There are a bunch of list of items floating around, and they have some overlap which isn't accounted for. For instance, switching from Craft:gear to set Recipe will still reset to accumulator.

Performance is **massively** improve by both only updating the element if is different and by using wxComboBox.Append with a wxArrayString instead of individual wxString's.

### DIFF
--- a/src/TypePanel.cpp
+++ b/src/TypePanel.cpp
@@ -151,6 +151,34 @@ string cMain::ExtractStep()
 
 	return "not found";
 }
+
+void cMain::UpdateCmbItem(wxArrayString* new_list)
+{
+	if (new_list != current.cmb_item)
+	{
+		current.cmb_item = new_list;
+		cmb_item->Clear();
+		cmb_item->Append(*new_list);
+		cmb_item->SetValue(*new_list->begin());
+		cmb_item->AutoComplete(*new_list);
+	}
+}
+void cMain::UpdateLabelItem(const wxString* new_text)
+{
+	if (new_text != current.label_item)
+	{
+		current.label_item = new_text;
+		label_item->SetLabelText(*new_text);
+	}
+}
+void cMain::UpdateLabelFromInto(const wxString* new_text)
+{
+	if (new_text != current.label_from_into)
+	{
+		current.label_from_into = new_text;
+		label_from_into->SetLabelText(*new_text);
+	}
+}
 #pragma endregion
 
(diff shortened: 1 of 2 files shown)


src/TypePanel.cpp
```diff
@@ -175,55 +196,34 @@ void cMain::OnTakeChosen(wxCommandEvent& event)
...
-	cmb_item->Clear();
-	for (auto it = all_items.begin(); it < all_items.end(); it++)
-	{
-		cmb_item->Append(*it);
-	}
-	cmb_item->SetValue(*all_items.begin());
-	cmb_item->AutoComplete(item_choices);
+	UpdateCmbItem(&item_choices);
+	UpdateLabelItem(&TypePanel::item);
+	UpdateLabelFromInto(&TypePanel::from);
```
MortenTobiasNielsen, line 214: I guess this should be _&TypePanel::into_ instead of _&TypePanel::from_

src/TypePanel.cpp
```diff
@@ -238,15 +238,8 @@ void cMain::OnfilterChosen(wxCommandEvent& event)
...
-	cmb_item->Clear();
-	for (auto it = all_items.begin(); it < all_items.end(); it++)
-	{
-		cmb_item->Append(*it);
-	}
-	cmb_item->SetValue(*all_items.begin());
-	cmb_item->AutoComplete(item_choices);
-
-	label_item->SetLabelText(TypePanel::item);
+	UpdateCmbItem(&handcrafted_choices);
```
MortenTobiasNielsen, line 241: I think this should be _&item_choices_

MortenTobiasNielsen (changes requested):
Nice work - it has been annoying that it changes the item when just changing the step type. It would be nice it was checked if the new list also has the item. So it keeps _assembling machine_ if I have just taken an assembling machine, I change it to build and then change it to recipe.

src/TypePanel.cpp
theis999: Correct

PR #183, opened 2023-03-07
Fix to check recipe and solid fuel recipe names

theis999:
Fix to recipe check for Oil refinery, Chemical plant, Centrifuge and Furnaces
Fix to solid fuel recipe names

I dont think there is much point in checking the recipe of a furnace is valid as you usually can't set it.

Solid fuel recipe names are very very long. We could shorten them and use a translation table to convert them
image


MortenTobiasNielsen:
Nice that you found the issues (and fixed them). 

I have set the recipe for furnaces multiple times, so I would like to keep the check. 

I do like to use the names from the game so the user isn't confused what items are. I would though be nice if the combo-box was a little larger, so the entire name could be shown.

MortenTobiasNielsen (changes requested):
Could you increase the size of the combo-box a little, so the name of long items can be seen fully?

theis999:
> I have set the recipe for furnaces multiple times, so I would like to keep the check.
Fair, I didn't think of that use. 

I managed to make it fit, by stealing 15px from the labels and also adding 15px on the right. "Solid fuel from petroleum gas" is an absurdly long name.

image

image

MortenTobiasNielsen (approved):

PR #184, opened 2023-03-08
Autocomplete to choose template

theis999:
Added autocomplete to choose template
Added onEnter event to choose template
Removed TemplateAlterStep as it wasn't implemented

### DIFF
--- a/src/GUI/GUI.fbp
+++ b/src/GUI/GUI.fbp
@@ -4765,7 +4765,7 @@
                                                             <property name="selection">-1</property>
                                                             <property name="show">1</property>
                                                             <property name="size"></property>
-                                                            <property name="style"></property>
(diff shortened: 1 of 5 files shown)


MortenTobiasNielsen (changes requested):
When I enter some text into the template combo-box and then click the one I want it doesn't change until I press enter, which is relatively unintuitive given that I am used to be able to click and it changes. Is it possible to have some kind of "on text change" or "on clicked" and check if that name exists and then select it?

theis999:
I changed the onTextEnter event to onText which does seem to work much better.

MortenTobiasNielsen (approved):

MortenTobiasNielsen:
It does work much better - thanks. :)

PR #198, opened 2023-03-08
Diff print

theis999:
Added write_state to appends some state variables to tas_state.txt
Allowing the user to make a crude diff of two different tas versions.

This is fairly crude attempt, but it might still be helpful @MortenTobiasNielsen 

It prints the state every 11 tick.
It resets the state file if you open the tas in MP.
It only prints the state if goal=debug

The state currently consist of the player's position and the contents of his inventory.


theis999:
Sample:
[tas_state.txt](https://github.com/MortenTobiasNielsen/Factorio-TAS-Generator/files/10920896/tas_state.txt)

PR #206, opened 2023-03-10
Select last selected rows on load

theis999:
Added selected rows to the bottom of tas file

Changed load tas file to:
 Re-Select last selected rows
 Double clicks the top row of last selected rows
 Scroll top row of last selected rows

It ignores selected blocks where either the top or bottom row is outside the scope of the step grid

image
As seen in the image, it re-select rows in grey instead of the usual blue. That wasn't intentional, but i like the effect.
It isn't evident from the image that it also scrolls the step grid to the top row. Nor is it visible that it loads the top row into Step Details and Step Type.

### DIFF
--- a/src/OpenTas.cpp
+++ b/src/OpenTas.cpp
@@ -58,6 +58,17 @@ open_file_return_data* OpenTas::Open(DialogProgressBar* dialog_progress_bar, std
 		return_data.success = true;
 	}
 
+	if (update_segment(file))
+	{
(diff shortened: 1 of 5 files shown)


MortenTobiasNielsen:
Is it possible to scroll down so the top row is maybe row 4 in the view? so instead of this:
image

It was this:
image

theis999:
Should be possible yes, it isn't a native functionality though.

MortenTobiasNielsen (approved):

PR #207, opened 2023-03-13
Added task tas interface: change step event

theis999:
Needed for the state printer tool.

### DIFF
--- a/Lua Files/control.lua
+++ b/Lua Files/control.lua
@@ -131,10 +131,15 @@ end
 
 ---@param by number
 local function change_step(by)
+	local task_ = 0
+	if steps and steps[step] and steps[step][1][1] then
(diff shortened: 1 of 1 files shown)


Lua Files/control.lua
```diff
@@ -131,10 +131,15 @@ end
 
 ---@param by number
 local function change_step(by)
+	local task_ = 0
```
MortenTobiasNielsen, line 134: I would normally place the underscore to indicate private variables in the front of the name instead of in the front. Is this the normal way of doing it in LUA?

theis999: I don't know if the custom of lua, my linter gave me a warning for redefining the name task. So i just added an underscore.

MortenTobiasNielsen (approved):

Issue #208, opened 2023-03-13
Reload script file button


theis999:
The tas helper mod by Pallet12 is very convenient, but the export is clunky. We could help this by having a button that reloads the current script file.

MortenTobiasNielsen:
That would be nice. Could maybe have a F5 shortcut. :)

theis999:
Good idea!

theis999:
Unless it would be better to have an input box where the user can write strings directly to end of the step grid.

MortenTobiasNielsen:
That would be pretty awesome, if there was an import steps button. So the user could just take the steps generated by the recorder tool and put it directly into the tool. Great idea. :)

theis999:
A simple mock-up:
![image](https://user-images.githubusercontent.com/18309265/225036004-45585fc5-8f1e-4711-bd5d-37a055c16912.png)

Does this match what you intended?

MortenTobiasNielsen:
Yeah something like that would be nice.

Might be nice if it is possible to add the steps to a template, which already exists.

PR #211, opened 2023-03-19
Import step

theis999:
### #208 Import Step

Added a panel to import steps
It has 2 methods of importing steps, into a template or directly into the steps grid.

#### Import into Steps
This uses a number spinner to place the imported steps an off-set either from the start or the end of the list.
It is not possible to import steps if the spinner is out of range

#### Import into Template 
This creates a new template and populates it with the imported steps.
It names the template based on a text input that needs to be valid before steps can be imported.

#### Validation
The template name and steps offset both have a validation. That will color them red if they are invalid.
There is also a cursory validation on the main import text field. That will disable the buttons, without coloring them.

### Other changes
I've imported wxWidgets.natvis file that should help with debugging. We can remove it if causes issues.
I've changed some of the file filters for better organization.


src/ImportStepsPanel.cpp
```diff
@@ -0,0 +1,322 @@
...
+	else if (row < 0 && -row <= rows+1)
+	{
+		import_steps_into_steps_ctrl->SetBackgroundColour(wxColour("Window"));
+		import_steps_into_steps_btn->Enable(!import_steps_text_import->IsEmpty());
+	}
+	else
+	{
+		import_steps_into_steps_ctrl->SetBackgroundColour(wxColour("Red"));
+		import_steps_into_steps_btn->Enable(false);
+	}
```
MortenTobiasNielsen, line 201: It seems like this part could be given its own method, so there isn't duplicates of the code in line 161 to 177..

src/ImportStepsPanel.cpp
```diff
@@ -0,0 +1,322 @@
...
+	StepGridData.insert(it1, step_parameters.begin(), step_parameters.end());
+
+	if (import_steps_clear_checkbox->IsChecked()) import_steps_text_import->Clear();
+
+	no_changes = false;
+
+	event.Skip();
+}
+
+bool cMain::validateTemplateName()
```
MortenTobiasNielsen, line 253: Can you walk me though what this is doing. I haven't worked with the GetValidator and TransferFromWindow, and where does import_steps_into_template_ctrl_validator comes from - well GUI_base.h, but you know how did it get there and what does it do? :)

MortenTobiasNielsen (changes requested):
This is very well thought out from a user perspective and the code seems pretty good as well - impressive. 👍  

I haven't tried to break it, but it seems to work as intended and the validation seems pretty good. I think most users will use it as intended - copy/paste and insert. So hopefully users won't find some crazy edge case which will break something. :)

theis999:
I am glad you like it. The validation of user input should be solid. Except the imported steps, which follows the logic of open tas. As such it is lenient or broken in the same way.

So you don't need to search for edge cases, since we already know some of them. For instance it is possible to specify invalid recipes for buildings, and other things that are specific to the building type.

MortenTobiasNielsen (approved):

PR #215, opened 2023-03-20
Tas interface timestamp

theis999:
Added some extra variables (generator_name, generator_version, tas_name, tas_timestamp) to the TAS remote interface.

I am not sure we need any of it, besides @MortenTobiasNielsen desire for a timestamp.


MortenTobiasNielsen:
I usually find that when things become possible it opens doors not even visible before, so maybe having time and a version will make something possible. :D

theis999:
I have updated the tas state printer and verified it works with or without this update.

MortenTobiasNielsen:
The changes will be included in #226

PR #216, opened 2023-03-20
Refactor step names

theis999:
Removed the extra list of step names.

Found this neat regex to help:
image

Note: this builds on the pr #211

PR #217, opened 2023-03-20
Refactor utils

theis999:
Moved orientation, priority and inventory to their own files
Moved a bunch of constants only related to opening the tas save file, to their own file

Based on #211 -> #216

### DIFF
--- a/src/ImportStepsPanel.cpp
+++ b/src/ImportStepsPanel.cpp
@@ -59,9 +59,9 @@ bool ImportStepsPanel::extract_steps(wxString steps, vector<StepParameters>& ste
 	while (update_segment())
 	{
 		counter++;
-		if (segments.size() != step_segment_size &&
-			segments.size() != step_segment_size_without_colour &&
-			segments.size() != step_segment_size_without_comment_and_colour)
+		if (segments.size() != tas_file::step_segment_size &&
+			segments.size() != tas_file::step_segment_size_without_colour &&
+			segments.size() != tas_file::step_segment_size_without_comment_and_colour)
 		{
 			wxMessageBox("It was not possible to read line ["+ std::to_string(counter) + "] as it only contained [" + std::to_string(segments.size()) + "] semicolons", "Text import error");
 			return false;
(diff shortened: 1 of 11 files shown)


src/ImportStepsPanel.cpp
```diff
@@ -78,8 +78,8 @@ bool ImportStepsPanel::extract_steps(wxString steps, vector<StepParameters>& ste
 		step.Direction = Capitalize(segments[6]);
 		step.Size = segments[7] != "" ? stoi(segments[7]) : 0;
 		step.Buildings = segments[8] != "" ? stoi(segments[8]) : 0;
-		step.Comment = segments.size() == step_segment_size || segments.size() == step_segment_size_without_colour ? segments[9] : "";
-		step.Colour = segments.size() == step_segment_size ? segments[10] : "";
+		step.Comment = segments.size() == tas_file::step_segment_size || segments.size() == tas_file::step_segment_size_without_colour ? segments[9] : "";
```
MortenTobiasNielsen, line 81: Why not use "Using namespace tas_file" here?

src/cMain.cpp
```diff
@@ -1209,19 +1209,19 @@ void cMain::Open(std::ifstream * file)
 	save_file_location = result->save_file_location;
 	generate_code_folder_location = result->generate_code_folder_location;
 
-	if (result->goal == goal_steelaxe_text)
+	if (result->goal == tas_file::goal_steelaxe_text)
```
MortenTobiasNielsen, line 1212: Why not use "Using namespace tas_file" here?

src/utils.h
```diff
@@ -4,6 +4,7 @@
 #include <vector>
 #include <map>
 
+#include "TasFileConstants.h"
 #include "Priority.h"
 #include "Orientation.h"
 #include "Inventory.h"
```
MortenTobiasNielsen: I wouldn't guess that utils need these includes - is it just cause it is easier than to figure out where the segments (Priority.h, etc.) is needed? :)

MortenTobiasNielsen (changes requested):

src/ImportStepsPanel.cpp
theis999: It is a good question, and there is some thought to it. I choose to specify that this is from the tas_file *stuff* since this file is otherwise unrelated to working with the tas file.

So since OpenTas and SaveTas are both obviously integrated with the tas file, they will use Using to shorten the constant names, while ImportStepsPanel in my opinion can specify that when needed.

src/cMain.cpp
theis999: I choose to specify that this is from the tas_file *stuff* since this file is otherwise unrelated to working with the tas file.

src/utils.h
theis999: I chose when i made it, that utils.h would be the header for all utility files. Basicly a collection of utilities. This could be a bad decision, but it made sense when i made it :-)

MortenTobiasNielsen:
Changes transferred to #228 due to some strange errors.

PR #218, opened 2023-03-20
Run from file-view

theis999:
Changed prefix of lua files allowing script generation while running file-view


MortenTobiasNielsen (approved):
Nice to get that fixed. :)

PR #219, opened 2023-03-21
Removed typepanel steptype

theis999:
Removed typepanel->STEP_TYPE->step_type as it was essentially a duplicated of StepType

Also removed a project reference to SearchUtilities.h which could not be located


MortenTobiasNielsen:
I am getting tons of errors like this, when I try to compile:
image

MortenTobiasNielsen:
Changes transferred to #228 due to some strange errors.

PR #225, opened 2023-03-24
Step subpanels

### LINKED ISSUE #224: Step type sub panels
It would be neat if the different action types were organized slightly more.
For instance this could be done by adding three sub panels:

- Character interactions
- Building interactions
- Game interactions

Before
image

After-mockup
image

Hopefully this would make it slightly faster to scan the action types for the correct one

theis999:
Categories the step types into 3 categories with a minor border between them


theis999:
image

I changed the border type to be more subtle and reduced list gap sizes: to make better room for more new types.

MortenTobiasNielsen (approved):

PR #226, opened 2023-03-28
Tas interface commands

### LINKED ISSUE #214: Timestamp
We should extend the interface with a generated timestamp.

While we are at, we can also include a generator version and a mod name.

theis999:
Added the TAS console commands to the interface, allowing other mods to use them.
Which is useful for my TAS helper, as i can create buttons that execute the commands.

Control.lua scrips of "Theis TAS helper" where it is used:
[control.zip](https://github.com/MortenTobiasNielsen/Factorio-TAS-Generator/files/11086984/control.zip)

Note: this is built on #215 to prevent merge conflicts

Closes #214

### DIFF
--- a/src/GenerateScript.cpp
+++ b/src/GenerateScript.cpp
@@ -21,13 +21,10 @@ void GenerateScript::reset()
 	y_building_size = 0.0f;
 }
 
-// Get current date/time, format is Www Mmm dd hh:mm:ss yyyy
+// Get current date/time, format is yyyy-mm-dd hh:mm:ss
 const std::string GenerateScript::currentDateTime()
 {
-	// https://stackoverflow.com/questions/997946/how-to-get-current-time-and-date-in-c
-	auto end = std::chrono::system_clock::now();
-	std::time_t end_time = std::chrono::system_clock::to_time_t(end);
-	return std::ctime(&end_time);
+	return std::format("{:%Y-%m-%d %H:%M:%S}", std::chrono::system_clock::now()).substr(0, 19);
 }
 
 void GenerateScript::ClearSteps()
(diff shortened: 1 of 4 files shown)


src/GenerateScript.cpp
```diff
@@ -18,15 +21,47 @@ void GenerateScript::reset()
...
 	total_steps = 1;
-	step_list = "local step = {}\n\n";
+	step_list = "";
+	std::stringstream ss (step_list);
+	ss << endl << "local tas_generator = {" << endl
+		<< tab << "name = \"" << generator_thumbprint.name << "\"" << comma_endl
+		<< tab << "version = \"" << generator_thumbprint.version << "\"" << comma_endl
+		<< tab << "tas = {" << endl
+		<< tab << tab << "name = \"" << name << "\"" << comma_endl
+		<< tab << tab << "timestamp = \"" << currentDateTime() << "\"" << comma_endl
```
MortenTobiasNielsen, line 51: I would like to use std::chrono::system_clock::now(); unless you really need the very specific date format made in the currentDateTime function.

src/GenerateScript.cpp
```diff
@@ -64,6 +99,7 @@ void GenerateScript::PaintWalk(string step, bool paint)
 
 void GenerateScript::generate(wxWindow* parent, DialogProgressBar* dialog_progress_bar, vector<StepParameters> steps, string& folder_location, bool auto_close, bool only_generate_script, string goal)
 {
+	this->name = folder_location.substr(folder_location.find_last_of('\\') + 1);
```
MortenTobiasNielsen, line 102: Is this used for anything?

MortenTobiasNielsen (changes requested):

src/GenerateScript.cpp
theis999: It is used exactly once, in the method GenerateScript::ClearSteps() line 50.

src/GenerateScript.cpp
theis999: Done

MortenTobiasNielsen (approved):

PR #228, opened 2023-04-01
Refactor transfer

MortenTobiasNielsen:
Transferred changes from PR #217 and PR #219 due to some strange errors in those PRs, which (for some reason) is not an issue in this branch.


theis999 (approved):
It looks reasonably

Issue #231, opened 2023-04-04
Cancel craft with 0 (all) value crashes Factorio


MortenTobiasNielsen:
It seems like it isn't expecting/handling -1 as a value for cancel all in the Lua code

theis999:
Apparently the count is required in [cancel_crafting](https://lua-api.factorio.com/latest/LuaControl.html#LuaControl.cancel_crafting) so that is why

PR #236, opened 2023-04-10
Changed lua cancel craft 'all' to cancel 1M items. 

theis999:
Which should be enough for all realistic scenerios. And prevents the game crashed report in #231


MortenTobiasNielsen:
It would be nice to get this aligned with the current develop branch. It seems like there are a lot more than in this branch than what I would expect. :)

theis999:
Oh that looks very weird. I'll try to fix it

theis999:
It should be better now.

MortenTobiasNielsen (approved):

PR #238, opened 2023-04-12
Added the rest of the top menu to shortcut changer

theis999:
Added the rest of the top menu to shortcut changer
Changed the setting type to use a map of mapped shortcuts changed the layout of sc to fit
Changed the layout of sc to fit the extra items

Validated that it works with changing a setting
Validated that it doesn't crash old setting file
Validated that it can save over old setting file

### DIFF
--- a/src/GUI/GUI.fbp
+++ b/src/GUI/GUI.fbp
@@ -709,7 +709,7 @@
                     </object>
                 </object>
             </object>
-            <object class="wxPanel" expanded="0">
+            <object class="wxPanel" expanded="1">
                 <property name="BottomDockable">0</property>
                 <property name="LeftDockable">0</property>
                 <property name="RightDockable">0</property>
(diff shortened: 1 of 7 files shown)


MortenTobiasNielsen:
The GUI is a little off:
image

Is it possible to get them more gathered at the top? I think it would be best if the spacing is something like the step detail panel:
image

theis999:
I have improved the GUI a little, reducing it's overall size and changed the sizer type to one that grows from the top.

MortenTobiasNielsen:
Is it possible to use more of the width? It seems unnecessary to have a lot of the shortcuts placed on multiple lines. I guess that is also what causes it to have the relatively large gap between the fields?
image

theis999:
I have given the labels a bit more room so GOTLAP is only on two lines, and right aligned the labels so they stick closer to their input box.

It seems you understand the fundamental problem, that there is a large range of text lengths. Fitting each without custom code for any of them, is tricky if not outright impossible. Presumably more menu points are going to be added too, which hopefully won't break the GUI too much.

Either way, i don't think this is an important GUI component. As the user will see it very very infrequent, so it just needs to work those few times.

MortenTobiasNielsen (approved):
Looking much better - thanks for making the change. :)

PR #239, opened 2023-04-12
Added method for printing waiting state

theis999:
## Core
The core idea, is when the TAS prints a warning it goes into warning_mode.
Then when a step is successfully executed it should terminate warning_mode.
Terminating warning_mode then prints a brief custom msg and a number of ticks that have elapsed since the mode began.

### Validation
I have validated that it works in most instances, but i haven't touched compatibility_mode. 
However the worst that can happen when a warning_mode termination is omitted, is that the next termination is inaccurate.

### localized strings
Factorio has something called localized strings which combines localization and string.format fairly neatly
On the lua side they work by `{"string_id", param1, param2}` where the id is used to specify the string to use and the params are substituted in somewhere specified by the string in locale.

### DIFF
--- a/Lua Files/control.lua
+++ b/Lua Files/control.lua
@@ -127,7 +127,15 @@ end
 --Print warning in case of errors in tas programming
 local function warning(msg)
     if debug_state then
+		global.warning_mode = global.warning_mode or {start = game.tick}
 		player.print(msg, {r=1, g=1}) -- print warnings in yellow
+	end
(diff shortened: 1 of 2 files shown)


MortenTobiasNielsen (approved):

MortenTobiasNielsen:
Sounds like a nice feature. I will merge it in and test it while I develop my run. :)

PR #240, opened 2023-04-13
Toggle search up or down

### LINKED ISSUE #229: Search up instead of down
I have found that most of the time, i am more interested in searching upwards in the step list. It could be possible to have both using right-click or shift+enter. It is however not trivial to change the search direction.

theis999:
Added checkbox to toggle search up/down
Moved search to the top row of the step panel, replacing empty space

### DIFF
--- a/src/GUI/GUI.fbp
+++ b/src/GUI/GUI.fbp
@@ -6435,8 +6435,8 @@
                 <object class="auinotebookpage" expanded="1">
                     <property name="bitmap"></property>
                     <property name="label">Steps</property>
-                    <property name="select">0</property>
-                    <object class="wxPanel" expanded="0">
+                    <property name="select">1</property>
+                    <object class="wxPanel" expanded="1">
                         <property name="BottomDockable">1</property>
                         <property name="LeftDockable">1</property>
                         <property name="RightDockable">1</property>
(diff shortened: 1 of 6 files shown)


MortenTobiasNielsen:
I haven't been able to open the GUI to check, but I am pretty sure that I am not gonna like the "replacing empty space" change, given that it most likely will move the buttons, which I have gotten quite used to. :)

I get this when I try to compile and open the application:
image

theis999:
It is not very different
image

You need to delete `generator-settings.json` as the new format of it is not backwards compatible. (or delete shortcuts parts)

MortenTobiasNielsen:
This might be a good time to add some kind of "use the old GUI layout" toggle thing - I don't really want this change, but understand that it does make sense to implement it. 

Isn't it possible to make it backwards compatible, so it checks if it is the old format and change it. I don't want to make a breaking change for something which seems relatively minor. I know it isn't a breaking change in terms of code, but if I need to write something like "ohh btw, you need to delete some stange file in the folder of your run to make this work" I consider it breaking for the user. :)

theis999:
You might have misunderstood me, it does upgrade the file correctly. At least in my test. The problem is that the old version of the generator can't use the new file format, and since you loaded up a version with the new format your settings file was upgraded. Then when you returned to a branch that doesn't use the new file format, the generator can no longer read the settings file.

I am not sure what you mean by old GUI layout?

MortenTobiasNielsen:
I see - that makes sense. 👍 

So it is possible to revert back to the GUI which is the current GUI ("old" if we implement this change).

theis999:
It sounds laborious to have a separate _old GUI_. Given how integrated the code is with the interface.

MortenTobiasNielsen:
Yeah. It would be pretty difficult to actually have two different GUIs, but in this case it is just the layout of the GUI components and it seems to be possible to change the layout either on startup or "on click".  It is done for some things in cMain:
image

It might be possible to also do it for the placement of components.

theis999:
I don't question that it is possible. 
Just how much extra work it would add in the future, and by extension who would do that work.

MortenTobiasNielsen:
Fair point. It looks better anyway - I will just have to get used to the new placement of the buttons. :)

MortenTobiasNielsen (approved):

PR #255, opened 2023-04-20
Added handler for "Never Stop" comment

MortenTobiasNielsen:
I need this for easier TAS development. It should be changed to be a modifier when we have that functionality.

### DIFF
--- a/Lua Files/control.lua
+++ b/Lua Files/control.lua
@@ -94,6 +94,7 @@ local function save_global()
 	global.tas.never_stop = never_stop
 	global.tas.use_all_ticks = use_all_ticks
 	global.tas.step_executed = step_executed
+	global.tas.not_same_step = not_same_step
 end
 
 --recreate crash site
@@ -1625,6 +1626,7 @@ local function create_tas_global_state()
 		duration = 0,
 		ticks_mining = 0,
 		idled = 0,
+		not_same_step = 1,
 	}
 end
 
@@ -1665,6 +1667,7 @@ local function migrate_global()
 	never_stop = global.tas.never_stop
 	use_all_ticks = global.tas.use_all_ticks
 	step_executed = global.tas.step_executed
+	not_same_step = global.tas.not_same_step
 
 	player = global.tas.player
 	if player then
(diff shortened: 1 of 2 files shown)


theis999 (approved):
Code looks fine

Lua Files/control.lua
```diff
@@ -5,6 +5,10 @@ local tas_generator = require("steps")
 local steps = tas_generator.steps
 local debug_state = require("goal")
 local run = true
+local never_stop = false
+local use_all_ticks = false
+local step_executed = false
+local not_same_step = 1
```
theis999, line 11: not_same_step does not appear to be in a global variable so it is not persistent across save/loads

Lua Files/control.lua
```diff
@@ -1414,7 +1459,24 @@ script.on_event(defines.events.on_tick, function(event)
 	if compatibility_mode then
 		backwards_compatibility()
 	else
+		step_executed = false
 		handle_tick()
+
+		if global.use_all_ticks_warning_mode and step_executed then
```
theis999, line 1466: step_executed is set to false 2 lines above, so this always evaluates to false

Lua Files/control.lua
```diff
@@ -1414,7 +1459,24 @@ script.on_event(defines.events.on_tick, function(event)
...
+			end_use_all_ticks_warning_mode()
+		end
+
+		if use_all_ticks and not step_executed and global.use_all_ticks_warning_mode == nil and not player.mining_state.mining then
```
theis999, line 1470: Same here, but inverse

Lua Files/control.lua
MortenTobiasNielsen: Step executed is changed in handle_tick (in the different branches, which has something to do with executing a step).

Lua Files/control.lua
MortenTobiasNielsen: Step executed is changed in handle_tick (in the different branches, which has something to do with executing a step).

Lua Files/control.lua
MortenTobiasNielsen: I think I have added it now - is it correctly done? :)

Lua Files/control.lua
theis999: I see. That wasn't obvious 👍

Lua Files/control.lua
theis999: It looks correct 👍

PR #256, opened 2023-04-20
Added mutual exclusive logic to radio btn back

theis999:
Custom implementation of mutual exclusive logic.

It has some funky behavior if you click on the panel instead of one the buttons.


MortenTobiasNielsen:
Yeah. That is pretty strange behavior when the panels are clicked. Do you have an idea why that happens?

theis999:
Not sure, it seems like a bug in wxWidget. Where the framework is a bit too overeager to match an event to a component

MortenTobiasNielsen (approved):

PR #257, opened 2023-04-24
Added override to build and recipe. 

MortenTobiasNielsen:
Also added default values to migrate_global to prevent the user to have to create a new game


theis999 (approved):
Seems sensibel.
I suppose Override should be a step modifier too

MortenTobiasNielsen:
Yeah - Override should be a step modifier.

PR #258, opened 2023-04-25
Step Modifiers

### LINKED ISSUE #245: Warning on change recipe
When setting a new recipe on an assembler, if it is crafting something it will empty the assembler and cancel the craft. However most of the time you want it to finish crafting that item before switching.
It might even be desirable to block recipe changes before it finalize the craft

### LINKED ISSUE #248: Step modifier column in steps grid
We currently use the comment field for some special modifying behaviors which is a rather obscure way to handle such things.

### Current
We currently uses the comment field for a fairly wide array of step modifying behavior:
- check-distance: "old"/"new" which goes into compatibility mode 
- check-distance: "Override" which overrides the building interaction calculation
- Pick up: "COMP:" which does some special compatibility for pickup
- Tech: "Cancel" which cancels the current research queue before adding the new technologi to it.

then we also
- Save: comment is used for filename

and
- Mining: "Split" i am not sure what this does

### Step modifications
This is in response to having **Set recipe** that blocks until the assembler is not crafting.
However a few other uses also comes to mind:
1. **No order** - execute task in no particular order, for instance if you want to take items from a bunch of furnaces and just want it done as fast possible.
2. **Wait for crafting** - When you have queued up intermediate items for crafting the next item, then it could be useful to **block&warn** before adding a new item to the crafting queue.
3. **Walk to** - specifically for walking to a far away item to execute some task (mining it or take items from a wreck). Then we could cancel the walk.

theis999:
### Step modifiers changes how a single step is interpreted
This is either in the Generator or in the TAS execution.

There are included 4 modifiers:
1. No order
2. Wait for
3. Cancel
4. Walk towards

**Wait for** and **cancel others** are fairly simple.
### No order
Is only applied to a small subset of action types, it also isn't very performant. So ideally the user will apply it while developing and then change the order of their commands to match the execution.

#### It works by
Setting up a step block to execute, by iterating from the current step and forwards until it finds a step that isn't "no order"
Then it runs through each step every tick.
If a step is executed, it will then be removed from the step block
When the step block is empty, it returns to normal operations.

#### Notes
Every executed step will print a debug message, so you can see the order they are executed in.
If the execution of the step block exceeds 25 times the number of steps, it will fail. This might be low?
To suppress warnings generated by steps that can't be executed, the warning function is overwritten while block is being executed.

### Walk towards
Can be used to generate a walk step where the direction matters more than the destination. It means that the TAS will proceed with the next walk step once every step in between has been executed.
This however does mean that the generator, loses the ability to know where the character is.

I have added the walk towards modifier on intermediate steps. As it made sense to me that for these steps the user doesn't care about the location of the character. This is however **not backwards compatible**, so it might not be a great idea?


MortenTobiasNielsen:
The GUI isn't scaled correctly and can be a little difficult to see what modifiers are on a step. 
image

theis999:
I restyled it so it fits better.
Added the usual double click functionality
And added click handler similar to the painter to "no order" so it's possible to change multiple steps.

I don't know how to display the potentially very long modifier strings, other than this.

src/ImportStepsPanel.cpp
```diff
@@ -72,12 +72,12 @@ bool ImportStepsPanel::extract_steps(wxString steps, vector<StepParameters>& ste
...
-		step.Orientation = Capitalize(segments[5]);
-		step.Direction = Capitalize(segments[6]);
-		step.Size = segments[7] != "" ? stoi(segments[7]) : 0;
-		step.Buildings = segments[8] != "" ? stoi(segments[8]) : 0;
+		step.Amount = segments.size() >= 3 && segments[3] != "" ? to_string(stoi(segments[3])) : "";
+		step.Item = segments.size() >= 4 && segments[4] != "" ? Capitalize(segments[4], true) : "";
+		step.Orientation = segments.size() >= 5 && segments[5] != "" ? Capitalize(segments[5]) : "";
+		step.Direction = segments.size() >= 6 && segments[6] != "" ? Capitalize(segments[6]) : "";
+		step.Size = segments[7] != "" ? stoi(segments[7]) : 1;
+		step.Buildings = segments[8] != "" ? stoi(segments[8]) : 1;
```
MortenTobiasNielsen: This seems to be inconsistent. Why do _direction_ need to check that the size is >= 6 when _Size_ doesn't?

Lua Files/control.lua
```diff
@@ -904,6 +918,13 @@ local function recipe()
 		return false;
 	end
 
+	if global.wait_for_recipe and player_selection.crafting_progress ~= 0 then
+		warning(string.format("Step: %s, Action: %s, Step: %d - Set recipe %s: The entity is still crafting.", task[1], task[2], step, item:gsub("-", " "):gsub("^%l", string.upper)))
+		step_reached = step
+		return false
+	end
+	global.wait_for_recipe = nil
+
```
MortenTobiasNielsen, line 927: I want to be able to specify an amount which should be finished in the assembler. I know I can go back and change what I put into the assembler, that can sometimes be 20-40 assemblers. It is therefore easier to just specify what you want finished in there, so you don't have to go back.

Lua Files/control.lua
```diff
@@ -915,15 +936,22 @@ local function recipe()
 end
 
 local function tech()
+	if global.cancel and player.force.current_research then
+		player.force.cancel_current_research()
+		return false
+	else
+		global.cancel = nil
+	end
+
```
MortenTobiasNielsen, line 945: Won't this consume a tick instead of adding and cancelling in the same tick? Will it consume multiple ticks?

Lua Files/control.lua
```diff
@@ -915,15 +936,22 @@ local function recipe()
...
 		player.force.research_queue = {}
 		player.force.add_research(item)
 		msg(string.format("Research: Cleared queue and %s added", item))
 		return true
 	end
 
-	player.force.add_research(item)
-	msg(string.format("Research: %s added", item))
+	local was_addded = player.force.add_research(item)
+	msg(string.format("Research: %s %s added", item, not was_addded and "not"))
```
MortenTobiasNielsen: I find this pretty hard to read - how does it behave?

Lua Files/control.lua
```diff
@@ -1154,6 +1184,54 @@ local function doStep(current_step)
...
+	}
+end
+
+local function execute_StepBlock()
+	local _success, _step_index, _step
+	for i = 1, #global.step_block do
+		_step_index = global.step_block[i]
+		_step = steps[_step_index]
+		_success = doStep(_step)
+		if _success then
```
MortenTobiasNielsen, line 1215: You can put doStep(_step) directly into the if statement, so you don't have to allocate memory for _success - it might be optimized away, but I am not sure whether LUA does that.

Lua Files/control.lua
```diff
@@ -1180,8 +1258,9 @@ local function handle_pretick()
 			msg(string.format("(%.2f, %.2f) Complete after %f seconds (%d ticks)", player_position.x, player_position.y, player.online_time / 60, player.online_time))
 			change_step(1)
 			debug_state = false
-		elseif(steps[step][2] == "walk" and walking.walking == false and idle < 1) then
+		elseif(steps[step][2] == "walk" and ( walking.walking == false or global.walk_towards_state) and idle < 1) then
```
MortenTobiasNielsen: Remove the space in front of walking.walking

Lua Files/control.lua
```diff
@@ -1267,7 +1348,35 @@ local function handle_ontick()
...
+			mining = mining + 1
+			if mining > 5 then
+				if player.character_mining_progress == 0 then
+					warning(string.format("Step: %s, Action: %s, Step: %s - Mine: Cannot reach resource", steps[step][1][1], steps[step][1][2], step))
+					debug_state = false
+				else
+					mining = 0
+				end
+			end
+		elseif steps[step][2] ~= "walk" and steps[step][2] ~= "idle" and steps[step][2] ~= "mine" then
```
MortenTobiasNielsen, line 1378: Why did you change the order of idle and mine?

src/cMain.cpp
```diff
@@ -1552,6 +1553,12 @@ void cMain::OnAddMenuSelected(wxCommandEvent& event)
 
 void cMain::UpdateParameters(GridEntry* gridEntry, wxCommandEvent& event, bool changeType)
 {
+	auto& m = gridEntry->Modifiers;
```
MortenTobiasNielsen: Change m to modifiers

src/cMain.cpp
```diff
@@ -1868,6 +1875,13 @@ StepParameters cMain::ExtractStepParameters()
 	stepParameters.PriorityOut = input_output[radio_output->GetSelection()];
 	stepParameters.Comment = txt_comment->GetValue().ToStdString();
 
+	string m = "";
```
MortenTobiasNielsen: Change m to modifiers

src/cMain.cpp
```diff
@@ -1868,6 +1875,13 @@ StepParameters cMain::ExtractStepParameters()
...
+	m += modifier_cancel_checkbox->IsEnabled() && modifier_cancel_checkbox->GetValue() ? "cancel, " : "";
+	m += modifier_no_order_checkbox->IsEnabled() && modifier_no_order_checkbox->GetValue() ? "no order, " : "";
+	m += modifier_wait_for_checkbox->IsEnabled() && modifier_wait_for_checkbox->GetValue() ? "wait for, " : "";
+	m += modifier_walk_towards_checkbox->IsEnabled() && modifier_walk_towards_checkbox->GetValue() ? "walk towards, " : "";
+	stepParameters.Modifiers = m.substr(0, m.size() > 2 ? m.size() - 2 : m.size());
```
MortenTobiasNielsen: I think this would be easier to read if it was:
```
stepParameters.Modifiers = modifiers.size < 2 ? "" : substr(0, m.size() - 2 );
```

src/cMain.cpp
```diff
@@ -2529,3 +2545,48 @@ void cMain::OnMainBookPageChanged(wxAuiNotebookEvent& event)
...
+	if (rows.size() < 2) return;
+
+	for (int row : rows)
+	{
+		StepType e = StepGridData.at(row).StepEnum; 
+		if (modifier_types.no_order.contains(e))
+			continue;
+		else
+			return;
+	}
```
MortenTobiasNielsen: It seems like this behavior could be confusing to the user - if I read it correctly. Won't it make it so that the behavior will depend on whether all the steps marked can have the "No Order" modifier, but it doesn't tell me that I am not able to do that with the currently selected steps. 

I think it would give a better user experience if "No Order" was disabled if the user has selected a set of steps which makes it impossible to use "No Order" or it only mark the steps which can get the modifier.

src/cMain.cpp
```diff
@@ -2529,3 +2545,48 @@ void cMain::OnMainBookPageChanged(wxAuiNotebookEvent& event)
...
+		for (int row : rows)
+		{
+			auto& step = StepGridData.at(row);
+			if (step.Modifiers.find("no order") != std::string::npos)
+			{
+				auto size = step.Modifiers.size();
+				auto index = step.Modifiers.find("no order, ");
+				step.Modifiers = index != std::string::npos ?
+					step.Modifiers.substr(0, index) + step.Modifiers.substr(index + 10, size - index - 10) :
+					step.Modifiers.substr(0, size - 8);
```
MortenTobiasNielsen: You should specify these _magic numbers_ (10 and 8)

MortenTobiasNielsen (changes requested):

src/ImportStepsPanel.cpp
theis999: Fair point, i do think it is the later parts like _Size_ that should follow.

Lua Files/control.lua
theis999: Yes, that is the idea. That it follows the rule that actions cost ticks. So it cost 1 tick per research to remove, however since dependent techs will be cancelled too it might not cost 7 ticks to cancel 7 technologies.

Lua Files/control.lua
theis999: I am uncertain how best to add such a value. As the _Amount_ field is already used (or abused). I can see 3 options:
1. Move the auto-fill amount to the auto-fill panel. Which opens the amount field for set recipe.
2. Add a Wait-for amount somewhere in the modifier panel. 
3. Change wait-for to only block when recipe progress is < 90%.

Lua Files/control.lua
theis999: I doubt LUA does much of anything _smart_, it is a very small language. Performance-wise I don't think it matters either way, but it is slightly easier to step through with a debugger this way.

Lua Files/control.lua
theis999: Not sure. I moved lots of lines around so it might just have been an accident.

src/cMain.cpp
theis999: Correct!

src/cMain.cpp
theis999: The UX of using no-order needs a lot more work. Getting everything to work with just one button is tricky. Some your idea might be useful.

Lua Files/control.lua
theis999: Now the code is wrong, and should have been: 
`msg(string.format("Research: %s%s added", item, not was_addded and " not" or ""))`

It is the same as `!was_added ? " not" : ""`

theis999: I have made the fix mentioned.

src/cMain.cpp
theis999: I have updated this, so it both colours the no-order checkbox indicating special behavior, and provides a warning telling the user if a step can't be updated this way.

theis999:
I have made several changes based on the comments.

Everything but one issue should be resolved now. The last one being wait-for, where i have requested input.

Lua Files/control.lua
MortenTobiasNielsen: Fair point - it most likely won't make a difference, so if it makes it easier to debug it probably worth it.

Lua Files/control.lua
MortenTobiasNielsen: I think we should do option 2 and put it into the modifier field, so it is "wait for 5" if you want to wait for 5 units. This way we don't introduce a breaking change by moving/changing the "meaning" of amount.

Lua Files/control.lua
MortenTobiasNielsen: I guess that makes sense, but it is also a breaking change. It might be nice to add a little remark in the wait for explanation that it can take multiple steps to execute. 

I guess it isn't a problem as long as the possibility to use the "Cancel" comment for tech is still there.

Lua Files/control.lua
MortenTobiasNielsen: Can you change it back - now that we any way need to make some changes? I am sure that it doesn't mean anything, but I don't really like changes without any meaning. :)

MortenTobiasNielsen (changes requested):
I have found these things when I was trying out the new functionality. 
Small typo in the msg box - should be "unable to **be** assigned"
image

These is not enough space in the left panel, which means that a horizontal scroll bar is added when the vertical scroll bar is added:
image

After playing around with the No order modifier I have somehow gotten it into a state where it keeps being red even though I should be able to add the modifier to the selected steps:
image

In this bugged state it seems to disables all modifier if only one step is selected:
image

When I selected build it stopped being disabled, but it seems to think that one or multiple of the selected steps have the No order modifier and doesn't seem to be aware that the game speed cannot get the No order modifier:
image

It would be nice to have some sort of explanation as to what blue and red means. Red somewhat makes sense - but why isn't it disabled if I cannot click it? I don't really understand blue - does it just indicate that more than one step has been selected?

It seems like the movement color indicators haven't been moved:
image

I find it pretty difficult to try out every possible combination of the modifiers, but I hope I have found most of what isn't working as intended.

Lua Files/control.lua
```diff
@@ -915,15 +936,22 @@ local function recipe()
...
 		player.force.research_queue = {}
 		player.force.add_research(item)
 		msg(string.format("Research: Cleared queue and %s added", item))
 		return true
 	end
 
-	player.force.add_research(item)
-	msg(string.format("Research: %s added", item))
+	local was_addded = player.force.add_research(item)
+	msg(string.format("Research: %s%s added", item, not was_addded and " not" or ""))
```
MortenTobiasNielsen: I still don't know how to read this. Can you write this instead if it provides the same result? It makes it so much easier to read for people which isn't neck deep in LUA code. :D
```
msg(string.format("Research: %s%s added", item, was_addded ? "" : " not"))
```

Lua Files/control.lua
theis999: Added it as separate issue

Lua Files/control.lua
theis999: Sure! that is cleaner.

theis999: Lua doesn't have `?` operator but we can use `and` to mimic the behavior.
But the order of conditions are important as the last item is the one returned.
So ``` "" and was_added or " not" ``` would result in `false` or `" not"`.

theis999:
I corrected the **typo** and reversed the **order of operators** for **tech** was added to simplify the expression.
I moved colouring related to walk to comment column.

Unfortunately I can't see the issue with the **scrollbars**. I was very careful with placing it **pixel perfect**. Maybe we have different monitor resolutions? anyway, as i can't see the issue you have and i want to use the full space. The easiest solution is for you to make the necessary adjustment.

#### No-order
It seems to me that the colouring of the no-order checkbox has you more confused. Which is a shame because it means that i have replaced one UI problem with another.

Anyway, it has 4 states.
1. Disabled - The current step-**type** doesn't allow no-order
2. Enabled (black) - The current step-**type** allows no-order
3. Blue - You have selected multiple rows **and** the current step-**type** allows no-order
4. Red - You have selected multiple rows **and** the current step-**type** doesn't allow no-order

In other words the colouring of the checkbox indicates whether the change will affect the step you currently designing in the top row, and has nothing to do with the selected steps in the step panel being able to be changed.

Whether the multiple steps selected can be changed is handled, when you try, using the warning box.

Lua Files/control.lua
MortenTobiasNielsen: Sad that it isn't possible, but sure lets go with that.

MortenTobiasNielsen:
Yeah that is really confusing - borderline misleading. I do understand that it is pretty much impossible to convey all of the information necessary  with just one checkbox and some colors, but it is almost better with no colors, in that case it at least doesn't give visual ques which can be interpreted wrong. 

Maybe we can have a No Order and a No Order Row or something like that, so it is clear that one affects the row and the other the current step?
I do find it a little more intuitive that it give the modifier to every step which can get it instead of giving me an error - but would that not make sense due to how the "mechanic" works?

Lua Files/control.lua
theis999: Yeah it is a bit sad with Lua's limitations but at least it isn't Udon.

theis999:
It sounds to me as if you see it as a button that looks like checkbox, while it is in fact a checkbox that also has a button like behavior. So when the checkbox changes color it indicates (1) that it has button-like behavior and (2) the color indicates whether it retains checkbox-like behavior, with red by no and blue being yes. 

Maybe an more intuitive solution is to use a button for the button-type behavior. The button could be placed at the same location as the checkbox - toggling between visibility of the two control elements based on how many rows the user selects.

I would find it highly misleading if it just gave the modifier to any step that is able to have it. As it would create 2 step blocks even though the user selected a continues block of rows. It would be fine for something like `Ctrl+click`, but the current setup can't detect that.

MortenTobiasNielsen:
Agreed - a button would be better. Sounds like a good idea to have it as a ctrl-click.

theis999:
I have swapped the logic to use a button instead of only relying on the checkbox.
The button looks slightly weird, as i had make it slightly shorter than standard to fit in checkbox grid layout.

I added a right click handler that doesn't give warnings.

MortenTobiasNielsen (approved):
Lets give it a try. :D

PR #261, opened 2023-05-02
State modifiers

### LINKED ISSUE #260: TAS state modifiers
The basic concept is to have one or more step types that toggles how the TAS behaves. Mainly for use in debugging or developing the TAS.

### Proposed Types
- Use All Ticks: A warning will be raised if no step was executed in a tick. This is very useful in very large builds, where you need to use very single tick.
- Never Stop: A warning will be raised if the character isn't walking. At a certain point in an any% run you never want the character to stand still, but it can be extremely time consuming to go through every single tick an ensure that the character is not standing still for a tick.
- Always stone path: A warning will be raised if the character isn't walking on stone path.
- Always handcraft: A warning when the crafting queue empties.
- Msg level: Debug is currently tied to the goal, which is not optimal. Further it would further be nice to have 2-4 levels such as Verbose, Debug, Warn & Info.

### Gui
It makes sense place these in the game panel
image

### Implementation
Obviously this is many tasks combined. But some of these already have some amount implementation.

Each of these will require a modest change to the step details panel.
They should further raise a TAS state change event for other mods.

theis999:
## State modifier warnings 
Added 4 state modifiers that enable specific warnings

image

### Each is very similar, with the condition that triggers them being the most different.
They are each toggling. So step x enables them, and step x+y disables them.
They are toggled in the pre-tick,
The condition of them is checked in post-tick
They each use the same function to print the warnings to the player

I put the names of them into a table so it is easier to manage.
I also used localized strings for the string messages.

### This is built on #258

### DIFF
--- a/Lua Files/control.lua
+++ b/Lua Files/control.lua
@@ -165,11 +165,27 @@ local function end_use_all_ticks_warning_mode()
 	end
 end
 
+local warnings = {
+	never_idle = "never idle",
+	keep_walking = "keep walking",
+	keep_on_path = "keep on path",
+	keep_crafting = "keep crafting",
+}
+
+local function end_state_warning_mode(warning, extra)
+	if debug_state and global[warning] then
+		game.print(
+			{"step-warning."..warning, steps[global[warning].step][1][1], game.tick - global[warning].start, extra}
+		)
+		global[warning] = nil
+	end
+end
+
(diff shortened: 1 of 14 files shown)


MortenTobiasNielsen (approved):
Seems like that should work. I will test it when I need them. :)

PR #267, opened 2023-05-11
More step modifiers

theis999:
### Added 3 new step modifiers
1. **Skip** - Completely new, but works as if the specific line has been commented out.
2. **Force** - This is the same as the more hacky "Override" comment.
3. **Split** - This is the same as the more hacky "Split" comment.

### Skip
Skip also has click handler attached that behaves similar to no-order. Though slightly simplified as there are no step types that can't be skipped.


MortenTobiasNielsen (approved):
Seems like it should work - I will test it as I need to use it.

PR #269, opened 2023-05-12
Shooting n throwing grenades

### LINKED ISSUE #41: Implement shooting
The tool won't really be complete before the user is able to shoot.

theis999:
Relative simple implementations of shooting and throwing grenades.


MortenTobiasNielsen (approved):
Seems like that should work. :)

PR #270, opened 2023-05-12
Clear console

### LINKED ISSUE #262: Onload - clear console
Having messages from previous segments is more annoying than useful.

[Player clear_console](https://lua-api.factorio.com/latest/LuaPlayer.html#LuaPlayer.clear_console)

theis999:


MortenTobiasNielsen:
I am not entirely sure I want this - can we make it a toggle like "Only Generate Step Files"? We might need a new menu to handle that type of modification. It could also be nice to have a "don't print out every time something is saved" toggle.

theis999:
I promise i will make it toggle-able when i am ready with another project.

MortenTobiasNielsen (approved):

PR #272, opened 2023-05-15
Undo and Redo

### LINKED ISSUE #99: Undo
This is a request from @CasparPetch (zaspar) to have an undo option in the TAS creator.

I expect that this is not trivial to implement or maintain, so I don't know where to place it on priority list.

I expect 
- We need a stack of actions the user does, that we can iterate back.
- Every action needs an opposite action, so we can undo it.

The last point is good to keep in mind, for any other change.

theis999:
This adds the option to Ctrl+Z to undo and Ctrl+Y to redo a few types of actions.

It is built State modifiers, so there are some unrelated commits included.

The change can be broken into 4 categories:
1. Undo and Redo has been added to the shortcut menu
2. Command stack, class has been added
3. Undo and Redo handlers has been added the can reverse the user actions
4. The clickhandlers for the commands have been modified to fit into the undo and redo handlers

### Command stack
The command stack is a custom circular buffer that can handle 32 elements. In such a way that overflow overrides old elements. However it is also possible to pop items of the stack (REDO) and further it is possible to pop_back items (UNDO) reusing the items on it.

### Not included
This only observes ADD, DELETE, CHANGE, MOVE UP, and MOVE DOWN for the steps grid.
Which means MOVE 5, import steps, colouring, no-order modifier, everything template, etc. Is not included.

### DIFF
--- a/src/cMain.cpp
+++ b/src/cMain.cpp
@@ -2907,5 +2907,50 @@ void cMain::OnUndoMenuSelected(wxCommandEvent& event)
 void cMain::OnRedoMenuSelected(wxCommandEvent& event)
 {
 	Command command = stack.PopBack();
+
+	switch (command.type)
+	{
+		case T_NULL:
+			break;
+		case T_ADD:
+			{
+				for (auto& [row, step] : command.rows)
+				{
+					AddStep(row, step);
+				}
+			}
+			break;
+		case T_DELETE:
+			{
+				wxArrayInt rows{};
+				for (auto& [row, _] : command.rows)
+				{
+					rows.Add(row);
+				}
+				DeleteSteps(rows, true);
+			}
+			break;
+		case T_MODIFY:
+			{
+				auto& [row, param] = command.rows[1];
+				ChangeStep(command.row, param);
+			}
+			break;
+		case T_MOVE_UP:
+			{
+				SelectRowsInGrid(command.rows);
+				MoveRow(grid_steps, true);
+			}
+			break;
+		case T_MOVE_DOWN:
+			{
+				SelectRowsInGrid(command.rows);
+				MoveRow(grid_steps, false);
+			}
+			break;
+
+		default:
+			break;
+	}
 }
 
--- a/src/cMain.cpp
+++ b/src/cMain.cpp
@@ -547,11 +547,12 @@ void cMain::OnAddStepClicked(wxCommandEvent& event)
 	}
 	else
 	{
-		change = AddStep(grid_steps->GetNumberRows(), ExtractStepParameters());
+		int row = grid_steps->GetNumberRows();
+		change = AddStep(row, ExtractStepParameters());
 		if (change.size() > 0)
 		{
 			stack.Push({
-				.row = *grid_steps->GetSelectedRows().begin() - 1,
+				.row = row,
 				.type = T_ADD,
 				.rows = change,
 			});
@@ -584,11 +585,12 @@ void cMain::OnAddStepRightClicked(wxMouseEvent& event)
 	}
 	else
 	{
-		change = AddStep(grid_steps->GetNumberRows(), ExtractStepParameters());
+		int row = grid_steps->GetNumberRows();
+		change = AddStep(row, ExtractStepParameters());
 		if (change.size() > 0)
 		{
 			stack.Push({
-				.row = grid_steps->GetNumberRows(),
+				.row = row,
 				.type = T_ADD,
 				.rows = change,
 			});
@@ -599,7 +601,7 @@ void cMain::OnAddStepRightClicked(wxMouseEvent& event)
 	event.Skip();
 }
 
-vector<tuple<int, StepParameters>> cMain::AddStep(int row, StepParameters stepParameters)
+vector<tuple<int, StepParameters>> cMain::AddStep(int row, StepParameters stepParameters, bool auto_put)
 {
 	vector<tuple<int, StepParameters>> returnValue;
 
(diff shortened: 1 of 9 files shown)


src/cMain.cpp
```diff
@@ -527,18 +527,35 @@ void cMain::UpdateTemplateGrid(wxGrid* grid, vector<StepParameters>& steps)
...
+		}
 	}
 	else
 	{
-		AddStep(grid_steps->GetNumberRows());
+		change = AddStep(grid_steps->GetNumberRows(), ExtractStepParameters());
+		if (change.size() > 0)
+		{
+			stack.Push({
+				.row = *grid_steps->GetSelectedRows().begin() - 1,
```
MortenTobiasNielsen: I get this access violation error crashing the tool. It seems to be an off by 1 error where it doesn't take into account that there might not be anything in the grid:
image

MortenTobiasNielsen (changes requested):
I wanted to test out the new functionality and tried to delete 3 random rows and press undo, but for some reason it 4 rows comes back. 
image
image
Not really sure why that would happen, it isn’t duplicating one of the rows. I thought for a second it might have taken row 10707 for some reason, but the coordinates are not the same.

I cannot seem to find the change, but it seems like the Save (file) shortcut has been overridden:
image

I have this setup for testing:
image
I first added the build step and then added the mine steps from 1 to 10 and then I deleted the build step. When I undo and redo multiple times, I sometimes get this: 
image
Where it seems like the tool thinks that it is mining the inserter even though it has not been built. It is not consequent about which row gets it. 
image

I changed 4 to 100 and 7 to 200 and then undid the changes, but for some reason 7 turned into 6. It did though turn 4 back to 4. 
image

theis999:
Your save shortcut can be set through the shortcut changer. Ctrl+S is just the default.

src/cMain.cpp
theis999: I'll be honest i probably didn't test no-selection properly. As it is so rare to get into that state, you can use ctrl+click to get into it. 
I have fix for this issue.

theis999:
It looks like the first issue you found is related to auto-put. As the extra step matches the recipe. So i have made a fix for it.

I can't seem to replicate the issue with mining, even though your explanation is good. Perhaps it was fixed with one of the other changes?

MortenTobiasNielsen:
True it isn't often, but would be nice if new players were able to add their first step without the tool crashing. :D

Fair point - I am just pretty sure that I haven't changed the shortcut - is it because you have changed it on the branch and it spills over to my "config"?

Well spotted. I think you are right that it is due to the recipe. It could very well be a side effect solved by one of your changes.

src/cMain.cpp
```diff
@@ -675,102 +720,163 @@ void cMain::OnChangeStepClicked(wxCommandEvent& event)
...
+	wxArrayInt rows = grid_steps->GetSelectedRows();
+	if (wxMessageBox(
+		rows == 1 ? "Are you sure you want to delete this step?" : "Are you sure you want to delete these steps?",
+		rows == 1 ? "Delete step" : "Delete steps", 
+		wxICON_QUESTION | wxYES_NO, this) != wxYES)
 	{
 		return;
 	}
+	int startRow = rows.at(0);
+	auto steps = DeleteSteps(rows);
```
MortenTobiasNielsen, line 773: I can see that this is populated with data, but I have no idea why it does that, given that it doesn't seem like DeleteSteps explicitly return anything other than on line 808, which shouldn't be hit.

src/cMain.cpp
```diff
@@ -2737,3 +2849,112 @@ void cMain::OnSkipClicked(wxCommandEvent& event)
 		}
 	}
 }
+
+void cMain::SelectRowsInGrid(vector<tuple<int, StepParameters>> rows)
+{
+	grid_steps->ClearSelection();
+	for (auto& [r, _] : rows)
```
MortenTobiasNielsen: I would be come comfortable with for (auto& row : rows). I don't really think decomposing provides any benefit here - it might though also just be the lack of a meaningful name for "r".

src/cMain.cpp
```diff
@@ -2737,3 +2849,112 @@ void cMain::OnSkipClicked(wxCommandEvent& event)
...
+			SelectRowsInGrid(command.rows);
+			MoveRow(grid_steps, true);
+		}
+			break;
+
+		default:
+			break;
+	}
+}
+void cMain::OnRedoMenuSelected(wxCommandEvent& event)
```
MortenTobiasNielsen, line 2907: Add a new line to separate the functions.

MortenTobiasNielsen (changes requested):
I tried to delete all steps in my file and "undo", which crashed the tool: 
image

I unfortunately still get the error with the mine step thinking that it is mining something which hasn't been build yet. 
image

Steps to reproduce (I hope):

1. Add build inserter step.
2. Add each of the mining steps starting with 1. 
3. Delete the Inserter step.
4. Use undo to get the inserter step back.
5. Use undo 5-6 times to remove some of the mining steps.
6. Press redo fast multiple times until all mining steps are back.

src/cMain.cpp
theis999: C++ magic apparently 🤣 I have no idea how that worked either. 
I have added a return statement at the end of DeleteSteps so it is a bit more sensible.

src/cMain.cpp
theis999: Personally i don't mind the name to much when it is used so close to it's definition. But i will change r => row.

theis999:
I have added a stack reset on open file, as that left stack in an obviously wrong state. Which i think fixes the first issue you highlighted.
I can reproduce the other error. Correcting it may take a while.

src/cMain.cpp
```diff
@@ -2278,17 +2280,9 @@ void cMain::UpdateStepGrid(int row, StepParameters* stepParameters)
...
-
-		StepGridData.insert(it1, *stepParameters);
-	}
-	else
-	{
-		StepGridData.push_back(*stepParameters);
-	}
+	auto it1 = StepGridData.begin();
+	it1 += row;
+	StepGridData.insert(it1, *stepParameters);
```
MortenTobiasNielsen, line 2286: Are you sure this works? I would have guessed that I added the branch for a reason.

src/cMain.cpp
```diff
@@ -2853,9 +2847,9 @@ void cMain::OnSkipClicked(wxCommandEvent& event)
 void cMain::SelectRowsInGrid(vector<tuple<int, StepParameters>> rows)
 {
 	grid_steps->ClearSelection();
-	for (auto& [r, _] : rows)
+	for (auto& [row, _] : rows)
```
MortenTobiasNielsen: It is somewhat strange extracting something from the rows list isn't a row, but instead that object has a row property in it. I know it is mostly semantics, but it is somewhat confusing to read, given that it doesn't really make sense.

src/cMain.cpp
```diff
@@ -844,7 +846,7 @@ vector< tuple<int, StepParameters>> cMain::DeleteSteps(wxArrayInt steps, bool au
 
 		StepGridData.erase(iStart, iEnd);
 	}
```
MortenTobiasNielsen, line 848: Add a new line.

MortenTobiasNielsen (changes requested):
How does this resolve the "step thinks it can mine a building which haven't been build yet" problem?

src/cMain.cpp
theis999: I am certain. All this did was push a row to different point in StepGridData than the grid_steps.

The branching doesn't make sense anyway. Since we already have a row to insert at. 
Overriding that row in a method this far down in the call-stack is incredibly confusing.

src/cMain.cpp
theis999: I think i can see what you mean. It is not entirely obvious what each part of the data type represent:
vector is the list of rows
tuple is a concatenation of row index and row data
StepParameters is the row data

`for (auto& [index, data] : rows)` might be more semantically correct.

theis999:
It solves by correcting the underlying data structure StepGridData. The issue you found was just a symptom, of the gui not representing the data correctly while the real issue was that the underlying structure was updated incorrectly.

Because you deleted all steps you ended in a situation where no rows were selected in the steps grid. When redo'ing steps these steps would not exclusively be added at the end of the steps list. However as there was no selection, the redone step was added correctly in steps_grid but at the end of StepGridData.

So if you ran a generate steps you'd see that the steps were different from what was in the GUI.

src/cMain.cpp
MortenTobiasNielsen: Agreed, that makes more sense.

src/cMain.cpp
MortenTobiasNielsen: Fair point, seems like all the refactoring have made this obsolete and possibly incorrect. Just somewhat strange, that it haven't caused any problems for me, now that I have used to the tool pretty extensively.

MortenTobiasNielsen (changes requested):

src/cMain.cpp
MortenTobiasNielsen: Fair point, seems like all the refactoring have made this obsolete and possibly incorrect. Just somewhat strange, that it haven't caused any problems for me, now that I have used to the tool pretty extensively.

src/cMain.cpp
MortenTobiasNielsen: Agreed, that makes more sense.

src/cMain.cpp
theis999: It wouldn't cause any issues as long as there isn't a discrepancy between the row you interact on and the row that is selected. It is really only Undo/Redo and import-steps that would insert rows invariant of row selection.

MortenTobiasNielsen (approved):

PR #279, opened 2023-05-19
Changed Goal and Log level

theis999:
Added dependency on MOD: Speedrun_goals
Added log config to control print level
Added print config for more specific printing control
Added variables.lua file to hold other generated variables 
Removed goal files as goals are now controlled by Speedrun_goals

Goals and log-level now have separate menus. Where the goal is controlled by speedrun goals.

### Speedrun goals
Speedrun goal handles rock yield and always day.

settings-updates.lua controls the settings of speedrun goals, so we can set the settings how we want them.

### Variables
The variables.lua file allows us to insert the variables we want / need without messing with control.lua or steps.lua. As both files are sufficiently large already.

### Log level
I have included 3 levels of logging. So we can reduce the amount of information printed to console more precisely.
We can easily enough introduce new levels with this setup.

In addition more specific settings are introduced to control printing of comment, research queue and save game.

### DIFF
--- a/Lua Files/control.lua
+++ b/Lua Files/control.lua
@@ -1,9 +1,9 @@
 local util = require("util")
 local crash_site = require("crash-site")
-
-local tas_generator = require("steps")
-local steps = tas_generator.steps
-local debug_state = require("goal")
+require("goals")
+local tas_generator = require("variables")
+local steps = require("steps")
+local debug_state = true
 local run = true
 local never_stop = false
 local use_all_ticks = false
@@ -114,15 +114,27 @@ local on_player_created = function(event)
     player.get_main_inventory().sort_and_merge()
 end
 
---Print message intended for viewers
-local function msg(msg)
-    player.print(msg)
+---Print message intended for viewers
+---@param message LocalisedString
+---@param color Color | nil
+local function Message(message, color)
+    player.print(message, color or {})
 end
 
---Print debug message about what the tas is doing
-local function debug(msg, supress_info)
-	if debug_state then
-		player.print(msg)
+---Print commment message intended for viewers
+---@param message LocalisedString | nil
+local function Comment(message)
+	if PRINT_COMMENT and message and message ~= "" then
+		player.print(message)
+	end
+end
+
+---Print Debug message about what the tas is doing
+---@param message LocalisedString
+---@param supress_info boolean? Includes extra information in message
+local function Debug(message, supress_info)
+	if LOGLEVEL == 0 then
+		player.print(message)
         if not supress_info then
 			player.print(string.format(
 				"Seconds: %s, tick: %s, player position [%d, %d]",
@@ -135,23 +147,35 @@ local function debug(msg, supress_info)
 	end
 end
 
---Print warning in case of errors in tas programming
-local function warning(msg)
-    if debug_state then
+---Print warning in case of errors in tas programming
+---@param message LocalisedString
+---@param color Color | nil Message color or default yellow
+local function Warning(message, color)
+    if LOGLEVEL < 2 then
 		global.warning_mode = global.warning_mode or {start = game.tick}
-		player.print(msg, {r=1, g=1}) -- print warnings in yellow
+		player.print(message, color or {r=1, g=1,})
+	end
+end
+
+---Print warning in case of errors in tas programming
+---@param message LocalisedString
+---@param color Color | nil Message color or default red
+local function Error(message, color)
+    if LOGLEVEL < 3 then
+		global.warning_mode = global.warning_mode or {start = game.tick}
+		player.print(message, color or {r=1,})
 	end
 end
 
 local function end_warning_mode(msg)
-	if debug_state and global.warning_mode then
+	if global.warning_mode then
 		player.print({"step-warning.mode", msg, game.tick - global.warning_mode.start,}, {r=1, g=1}) -- print warnings in yellow
 		global.warning_mode = nil
 	end
 end
 
 local function end_never_stop_modifier_warning_mode()
-	if debug_state and global.never_stop_modifier_warning_mode then
+	if global.never_stop_modifier_warning_mode then
 		player.print(string.format("Step: %d - The character stood stil for %d tick(s) ", global.never_stop_modifier_warning_mode.step, game.tick - global.never_stop_modifier_warning_mode.start),{r=1, g=1})
 		global.never_stop_modifier_warning_mode = nil
 	end
@@ -159,7 +183,7 @@ end
 
 -- I think it should be possible to do something like looping through the different types and check if any are tagged for termination
 local function end_use_all_ticks_warning_mode()
-	if debug_state and global.use_all_ticks_warning_mode then
+	if global.use_all_ticks_warning_mode then
 		player.print(string.format("Step: %d - %d tick(s) not used", global.use_all_ticks_warning_mode.step, game.tick - global.use_all_ticks_warning_mode.start), {r=1, g=1})
 		global.use_all_ticks_warning_mode = nil
 	end
@@ -173,7 +197,7 @@ local warnings = {
 }
 
 local function end_state_warning_mode(warning, extra)
-	if debug_state and global[warning] then
+	if global[warning] then
 		game.print(
 			{"step-warning."..warning, steps[global[warning].step][1][1], game.tick - global[warning].start, extra}
 		)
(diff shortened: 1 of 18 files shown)


Lua Files/control.lua
```diff
@@ -115,13 +115,19 @@ local on_player_created = function(event)
 end
 
 --Print message intended for viewers
-local function msg(msg)
+local function Message(msg)
```
MortenTobiasNielsen: Good idea to use message instead of msg - can you make the change to the input as well? :)

Lua Files/control.lua
```diff
@@ -1,9 +1,9 @@
 local util = require("util")
 local crash_site = require("crash-site")
 
-local tas_generator = require("steps")
-local steps = tas_generator.steps
-local debug_state = require("goal")
+local tas_generator = require("variables")
+local steps = require("steps")
+local debug_state = LOGLEVEL < 2
```
MortenTobiasNielsen: Why is debug_state active when LOGLEVEL is below 2 when it needs to be LOGLEVEL == 0 for it to print a debug message? (line 130)?

Lua Files/control.lua
```diff
@@ -199,12 +205,12 @@ end
 local function save(task, nameOfSaveGame)
 	save_global()
 	if game.is_multiplayer() then
-		debug(string.format("Step: %s, saving game as %s", task, nameOfSaveGame))
+		if PRINT_SAVEGAME then debug(string.format("Step: %s, saving game as %s", task, nameOfSaveGame)) end
```
MortenTobiasNielsen: Why are LOGLEVEL, PRINT_SAVEGAME, etc. in all caps?

Lua Files/control.lua
```diff
@@ -115,13 +115,19 @@ local on_player_created = function(event)
 end
 
 --Print message intended for viewers
-local function msg(msg)
+local function Message(msg)
     player.print(msg)
 end
 
+local function Comment(msg)
```
MortenTobiasNielsen: Good idea to add a Comment function - cleans up the code nicely.

Lua Files/control.lua
```diff
@@ -1772,19 +1777,6 @@ script.on_event(defines.events.on_player_mined_entity, function(event)
...
-	end
-
-	if event.entity.name == "rock-big" then
-		-- do nothing, big rocks are always 20 stone
-	end
-
-	if event.entity.name == "sand-rock-big" then
-		mining_event_replace(event, "stone", 24)
-	end
-
```
MortenTobiasNielsen: I am not sure I like this idea that the tool is so tightly coupled with an external tool.

Lua Files/settings-updates.lua
```diff
@@ -0,0 +1,38 @@
...
+require("variables")
+
+local settings = {
+    ["t-sg-speedrun-goal"] = {type = "string-setting", value = GOAL}, --["Steelaxe%", "GOTLAP", "Default settings", "Any%", "Robot rush"]
+    ["t-sg-always-day"] = {type = "bool-setting", value = true},--[true,false]
+    ["t-sg-rock-rng"] = {type = "string-setting", value = "Static"},--["Default", "Custom", "Static"]
+    ["t-sg-rock-rng-seed"] = {type = "int-setting", value = 5},--[1 : 100]
+    ["t-sg-rock-huge-coal"] = {type = "int-setting", value = 47},--[24 : 50]
+    ["t-sg-rock-huge-stone"] = {type = "int-setting", value = 47},--[24 : 50]
+    ["t-sg-sand-rock-big-stone"] = {type = "int-setting", value = 24},--[19 : 25]
```
MortenTobiasNielsen: I very much don't like this idea - I think we have talked about it at a few times.

MortenTobiasNielsen (changes requested):
It would be nice if there was a little explanation of what each log level does - it is not really obvious what debug, development and release does.

Lua Files/control.lua
theis999: Done

Lua Files/control.lua
theis999: This might have been a stupid decision from me.
Previous we had 2 configurations debug and not_debug, while it is obvious that the new "Release" configuration match not_debug. It is however not obvious if debug matches the new "Debug" or "Development".

The best approach might be to just straight up remove debug_state entirely?

Lua Files/control.lua
theis999: It was meant to indicate that they are global variables. But it does follow the Lua style for constants.

Lua Files/control.lua
theis999: You do have mod privilegies so you can ensure that it is up to standard.
For release of a TAS the supporting mod is automatically downloaded & enabled, and for development it should be fairly easy for most people to ensure it is enabled.

Is there any job you would be fine with having a supporting mod take care of?

theis999:
It is a bit annoying that the help doesn't show up for log-levels as i did actually type it out.
image

**Speedrun Goals**
One of my projects for this mod is to create a custom GUI for winning the game. Which includes the total time taken for beating the game in milliseconds and time to research technologies instead of kills
image
As this is likely to include several lines of code to create, i want it far away from the rest of the TAS handling.

Lua Files/control.lua
MortenTobiasNielsen: I guess LOGLEVEL could be used instead. So when debug_state should be false we set the LOGLEVEL to 2?

Lua Files/control.lua
MortenTobiasNielsen: I appreciate that I have mod privileges, but I would still like this tool to be able to function without the user having to "remember" that a mod is needed for it to function. 

I think the idea of having a nice GUI for the winning condition makes a lot of sense and will enhance the use experience, and it doesn't need to be a part of the FTG tool, so it is a good idea to externalize it, like the other semi-support tools.

MortenTobiasNielsen:
Strange that the help message isn't shown, but seems to be an issue with the framework, so nothing really to do about it other than crossing fingers that it will be solved at some point. 

Sounds like a nice mod, which will be useful for many players. I did though think that speed runners aren't allowed to use mods, so will they be able to use this?

Lua Files/control.lua
theis999: I have revamped debug_state so it is exclusive to compatibility mode.

Lua Files/control.lua
theis999: It is not exactly something you can forget. As the TAS mod simply won't load without the dependency being enabled too.

Lua Files/settings-updates.lua
theis999: I've removed the reliance on Speedrun goals' rock handling.
So it is only handles ending the game.

theis999:
> ...I did though think that speed runners aren't allowed to use mods, so will they be able to use this?

They do, so they won't. But that doesn't particularly matter to me :)

I've changed the integration with the mod, so it only handles the endgame trigger and displaying the endgame gui.
I have removed most uses of debug_state.
I have added a new Error method, which should only be used for fatal errors.

Lua Files/control.lua
MortenTobiasNielsen: Fair, but it really is nicer that it just works, instead of it telling me that it doesn't work because I don't have a "third-party" mod installed.

MortenTobiasNielsen:
When I generate script with this branch my run doesn't work at all. It seems to just be something with it needing a dependency, but I would really appreciate it if that wasn't the case. I do not think the dependency provides enough value for it to warrant a breaking change.

Lua Files/control.lua
theis999: I am not going to pretend i understand the issue you have. If the developer distribute their TAS through the mod portal dependencies are automatically download & enabled. If the developer have created other TAS's then they should know what to do. If the developer switch between different TAS mods, the mod will automatically be enabled. When is this an issue?

Seeing that the mod is not loaded is simple enough:
image

theis999:
I have run my steel axe TAS using this branch, and it worked perfectly fine. Providing me with End game GUI that i previously had to hack in.

Lua Files/control.lua
MortenTobiasNielsen: How would the developer know that they suddenly have to install a mod to make it work - it haven't been needed before? I am aware that it is easy to do - like it is easy to download an extension, but I still would still like for a tool to at least function without external things. 

Why is it so important to tightly couple the goal tool instead of emitting an event like we do with other things?

theis999:
I'll close this PR and reopen a new one to keep it more focus as a lot have changed.

PR #282, opened 2023-05-22
Split shortcut menu into shortcuts and step types

theis999:
- Added new menu Step types
- Moved step types shortcuts to new menu.
- Added step types shortcuts for those that were missing. 
- Added new shortcut to select the search field.
- Added right click handlers to change step and delete step. 
- Added shortcut that match right click handlers of buttons. 
- Added new page in shortcut changer to handle the new menu.
- Added tooltips to buttons Step panels that didn't have them
- Fixed search columns, oversight with modifiers.

image
image
image
image

PR #285, opened 2023-05-23
Refactor update parameters

theis999:
- Removed building_orientation from StepType:Priority.
- Added lookup array matching StepType to parameter_choices
- Refactored cMain::UpdateParameters

**UpdateParameters** has grown way to large. This should help reduce it's size. I have used the system of parameter_choices to find out which control elements that should be updated.
I can reduce the size of the function further. However some of On___SelectMenu have more code in them so i would need to put the functions into a lookup table. I don't think it would be worth it currently.

It looks a bit silly to have a switch where all but one case seemingly does the same, but it should at least be easier to work with.

I removed building_orientation from Priority as the value isn't relevant. And the field is being used for priority. Can this cause issues?

This is built on #282 to get the select menu functions.

### DIFF
--- a/src/GUI/GUI.fbp
+++ b/src/GUI/GUI.fbp
@@ -81,7 +81,7 @@
                 <property name="window_extra_style"></property>
                 <property name="window_name"></property>
                 <property name="window_style"></property>
-                <object class="wxMenu" expanded="0">
+                <object class="wxMenu" expanded="1">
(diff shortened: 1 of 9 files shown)


MortenTobiasNielsen (approved):
Nice refactor. I am still finding it a little difficult to read/understand the bitwise operations, but it it neat none the less. :)

PR #292, opened 2023-06-02
Split debug from goals

theis999:
This is based on #279 
but has been remade to keep the discussion more focused.

### Commit details:
Added debug levels.
Split debug from goal, old debug will default to steel axe. Merged all goals into one file.
Changed "Any %" to "Any%".
Added variables file to insert other variable data that isn't steps. Moved generator_thumbprint to variables file.
Capitalized lua warning and debug function.

### Debug levels
Debug state has been split into 3: Debug, Development, Release. 
Which gives better control of how much should printed to the console.
Further settings for not printing comments, tech and saves have been added.
image

### Goal
The four goal files have been merged into a single file, that handles our 3 goals.
This allows better reuse of code, and it should also be faster to generate (marginally).

Steel axe and GOTLAP also triggers the endgame GUI built into the game.
The print statement has been changed to use a localized string
image

### The print functions debug and warning have been capitalized
This by effect un-hides the built-in debug function, that prints to the Factorio log.

### DIFF
--- a/Lua Files/control.lua
+++ b/Lua Files/control.lua
@@ -1,9 +1,9 @@
 local util = require("util")
 local crash_site = require("crash-site")
-
-local tas_generator = require("steps")
-local steps = tas_generator.steps
-local debug_state = require("goal")
+require("goals")
+local tas_generator = require("variables")
+local steps = require("steps")
+local debug_state = true
 local run = true
 local never_stop = false
 local use_all_ticks = false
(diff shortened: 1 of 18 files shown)


MortenTobiasNielsen (changes requested):
Is it supposed to print in black?
image

It seems like it isn't printing save when I have checked it.

theis999:
Both issues should be fixed now.

MortenTobiasNielsen (approved):
Seems to work now. 👍

PR #293, opened 2023-06-04
Steps focus mode

### LINKED ISSUE #287: Current segment panel
This is an idea based on my experimentation of implementing segments. Let's call it segments lite.

The idea is to have single panel, that is virtually a copy of the steps panel. With buttons and steps grid. The grid uses the same table base as grid_steps so they have identical content however this grid hides every row before a certain step. Which is the last save step.

In this type of implementation, OnFocus handlers will make switch between which grid is active. Because there is only one new panel it should be easy to handle.

The benefit is that the user has fewer rows to look at, the scrolling of the panel therefor becomes easier and searching for a step doesn't yield results several hundred steps ago.

A suggestion on the name would be appreciated.

theis999:
Added steps focus checkbox,
With clickhandler that trim the steps grid to those after the last save point.

### DIFF
--- a/src/GUI/GUI.fbp
+++ b/src/GUI/GUI.fbp
@@ -8233,6 +8233,72 @@
                                             </object>
                                         </object>
                                     </object>
+                                    <object class="sizeritem" expanded="1">
+                                        <property name="border">5</property>
+                                        <property name="flag">wxALL</property>
+                                        <property name="proportion">0</property>
+                                        <object class="wxCheckBox" expanded="1">
+                                            <property name="BottomDockable">1</property>
+                                            <property name="LeftDockable">1</property>
+                                            <property name="RightDockable">1</property>
+                                            <property name="TopDockable">1</property>
+                                            <property name="aui_layer"></property>
+                                            <property name="aui_name"></property>
+                                            <property name="aui_position"></property>
+                                            <property name="aui_row"></property>
+                                            <property name="best_size"></property>
+                                            <property name="bg"></property>
+                                            <property name="caption"></property>
+                                            <property name="caption_visible">1</property>
+                                            <property name="center_pane">0</property>
+                                            <property name="checked">0</property>
+                                            <property name="close_button">1</property>
+                                            <property name="context_help"></property>
+                                            <property name="context_menu">1</property>
+                                            <property name="default_pane">0</property>
+                                            <property name="dock">Dock</property>
+                                            <property name="dock_fixed">0</property>
+                                            <property name="docking">Left</property>
+                                            <property name="drag_accept_files">0</property>
+                                            <property name="enabled">1</property>
+                                            <property name="fg"></property>
+                                            <property name="floatable">1</property>
+                                            <property name="font"></property>
+                                            <property name="gripper">0</property>
+                                            <property name="hidden">0</property>
+                                            <property name="id">wxID_ANY</property>
+                                            <property name="label">Focus</property>
+                                            <property name="max_size"></property>
+                                            <property name="maximize_button">0</property>
+                                            <property name="maximum_size"></property>
+                                            <property name="min_size"></property>
+                                            <property name="minimize_button">0</property>
+                                            <property name="minimum_size"></property>
+                                            <property name="moveable">1</property>
+                                            <property name="name">steps_focus_checkbox</property>
+                                            <property name="pane_border">1</property>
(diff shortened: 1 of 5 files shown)


theis999:
I forgot that it needed to handle new save steps.

MortenTobiasNielsen:
It needs to handle the situation where the user doesn't actually have a save step. 
image

In my 45k step run it freezes the GUI for 30+ seconds when I press the checkbox.

theis999:
I found a way to improve the performance.
I have added handling of modifying the steps to also handle focus mode.

MortenTobiasNielsen:
- The performance is better, but it is still very noticeable that the UI freezes. It might be that it isn't possible to increase performance more, but then it might be nice to have some kind of "I am processing" indicator. 
- When I delete a step I get sent to the bottom of the grid - which is a little annoying if I am at step 9k out of 45k. 
- When I change a step it is doing a strange jump (seems to be down to the bottom and then up again to where I was)
- I was a little worried about performance in terms of when saves are removed or added, but that seems to be pretty much instant. 👍 

I still get the access violation. In this case when I had the focus checkbox checked, pressed New, and unchecked the focus checkbox. 
image

theis999:
I am unsure if performance can be improved further, since the thing that is slow is the UI update. The fundamental problem is that the wxGrid isn't great at handling 45k rows. If it was the background data, then i could look at threads, but since it is the UI locking it up is somewhat desirable. 
It appears that the performance of the grid is dominated by how many rows are shown, even while doing batched updates. This means that hiding the **first** 1k rows takes more than 10 times as long than the **last** 1k rows.

I have changed the goto row logic, so it works more inline with the rest of the app.

I have added intermediate GUI updates for the large update that happens when the focus checkbox is clicked. So it is more obvious that the app is processing. This is at least 25% slower, but it feels better.

I think the access violation problem is solved too.

MortenTobiasNielsen:
I am not really convinced about the "processing indicator" - it is a little strange to me to have it scrolling like that. I am honestly a pretty confused about the performance here. It seems like the performance is pretty much the same when it is doing all of this rendering, it seems like it should be noticeably slower than doing all of the heavy lifting in memory and then making one UI update.

theis999:
I don't understand why the performance is so bad either. I tried a multitude of things to improve it.

With **44.064** rows:
- It takes [2.05 - 2.20] seconds without batching.
- It takes [0.85 - 0.91] seconds with batching.
- It takes [0.89 - 0.93] using wxGrid::SetRowSizes
- It takes 10+ **minutes** to update the grid after every HideRow.
- It takes [2.95 - 3.10] seconds to clear the grid and re-insert every row.

Further, different batch size obviously matters. With the current batch size of 800, it takes about 1.03 seconds which is 15% slower than pure batching. With a batch size of 100, it takes nearly 2 seconds or 95% of **not** batching.

 In conclusion
Performance is bad regardless of what we do and exponentially worse the larger the grid. Batch size 200 is about the limit of when updating the UI becomes too performance heavy.

MortenTobiasNielsen:
Doesn't it take less 2-3 seconds to start up the tool and load the 44K steps? Do you allocate all of the rows in one go when you clear the grid and re-insert every row? 

The fact that the time penalty becomes less per row depending on how many rows it is makes me think that it is a memory allocation issue.

theis999:
I have deleted the code that cleared and reinserted, so this is from memory: I cleared the grid; then gave the grid a row size list; then inserted all the rows(empty); and then updated the data of every cell in the grid.

It might be faster than Open due to the use batching. Which optimizes some processes within the wx framework.

It is somewhat clear from my test, that the wxGrid does not work well with thousands of rows. It simply doesn't scale well, so any problem we see with 50k rows is going to be 4x worse with 100k rows. Realistically we have 2 paths improve this underlaying issue: Either we find another wx component that is better, or we can look at a custom implementation of a wxGrid.

As for the performance of this function, there is realistically only one option that is better than what we have here. Which is to have 2 grids: a full grid & a focused grid. So the performance cost can be amortized across all other eventhandlers. But is that worth it? how often will anyone switch focus on/off

MortenTobiasNielsen (approved):
A little concerning with the scaling, but I think you are right that users won't be switching back and forth on a regular basis. We might though want to give the user the ability to decide how many saves they want to see. 

I was also think about keeping two panels in memory, but the impact on the code would be pretty absurd.

PR #294, opened 2023-06-08
Change color of steps

### LINKED ISSUE #288: Choose step type colour
It would be neat to have a way to choose your own custom colours for step types. 

It could be added to the generator-settings file.
The GUI part could be done similar to the change shortcuts.

The immediate goal for me is to colour Walk & Mine. As these are just white, however i often use them as control steps.

theis999:
- Added steptype colour to handle painting steps grid and imporing colour settings.
- Added colour settings to generator settings.
- Added GUI dialog to change steptype colours.

### Added a GUI window to handle colours
image
image
image
image

### Based on Shortcut changer
This is largely based on the shortcut changer. With a few minor differences and some major upgrades.

1. I have elected to split the step types according to their type type (type^2) so they align with the Top GUI. However this split is GUI only, so the code deals with it in a custom way.
2. The step and template grids are updated immediately. It doesn't appear like it cost a lot of performance.
3. Similar to the shortcut changer, the generator settings file is only updated when the dialog is dismissed.

### Other
As per usual the generator file is forwards upgradable but not backwards.

I meant to include the colors of diagonal & straight walk, as well as intermediate walk. But this PR was already huge 😎

Also do you prefer color or colour? It seems like github, factorio and my browser prefers color. While wxWidget prefers colour.

### DIFF
--- a/src/Factorio-TAS-Generator.vcxproj
+++ b/src/Factorio-TAS-Generator.vcxproj
@@ -179,6 +179,7 @@
     <ClCompile Include="StepParameters.cpp" />
     <ClCompile Include="SearchUtil.cpp" />
     <ClCompile Include="StepNameToEnum.cpp" />
+    <ClCompile Include="SteptypeColour.cpp" />
     <ClCompile Include="TypePanel.cpp" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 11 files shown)


MortenTobiasNielsen (approved):
Nice work. Lets just go with colours for now, we can always change that.

PR #295, opened 2023-06-09
Paint state modifier 

### LINKED ISSUE #291: Paint state modifiers based on state
It is a bit hard to gauge if a state modifier (such as keep-walking) is on.

A simple method is to count the number of them, perhaps during script generation, and then paint the modifier green if there is an odd amount of that state among the steps.

theis999:
Paints never_idle, keep_walking, keep_on_path & keep_crafting green. 
When there is an uneven amount of those steps, to indicate it is turned on.

### DIFF
--- a/src/GenerateScript.cpp
+++ b/src/GenerateScript.cpp
@@ -12,6 +12,7 @@ GenerateScript::GenerateScript(wxGrid* grid_steps) : grid_steps(grid_steps)
 
 void GenerateScript::reset()
 {
+	warning_state_counters = WarningsStatesCounters();
 	ClearSteps();
 	player_x_cord = 0.0f;
 	player_y_cord = 0.0f;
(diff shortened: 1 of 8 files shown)


MortenTobiasNielsen (changes requested):
This implementation can be a little misleading/confusing, given that it doesn't know where in the grid I am. I might be doing something at step 13K out of 30K and the marker being green is telling me that at the end of the 30K it will be on, but not what it is at the 13K. 

Is it possible to mark it the same way we mark e.g. Diagonal?

theis999:
I understand what you mean, but I think it will be a very infrequent issue. Most of the time people won't go far back in step list to modify these types of steps. Since they mostly matter for active development.

Anyway, i will try just because i like having more colour in the GUI.

theis999:
Done

MortenTobiasNielsen (approved):
It would be nice if the steps were colored immediately instead of only being done when the script is generated, but we can have a look at that at a later time. :)

PR #298, opened 2023-06-16
Improved performance of open tas

theis999:
- Simplified open tas, extract steps.
- Improved performance loading steps in grid_steps by using batch.
- Improved performance of open tas by adding new Capitalize (std) function that doesn't convert back and forth between std and wx string.

### Test
I did some test on open TAS with 204k rows. In my test it took 37-39 seconds to load all those rows. The 2 steps that contributed the most was PopulateStepGrid (17s) and OpenTas::extract_steps (19s)

Using beginBatch & endBatch reduced PopulateStepGrid from ~17 seconds to ~10 seconds. So reducing that by 40%.

Changing the Capitalize function, so it didn't implicitly convert segments strings from std::string to wxString before converting them back std::string. Improved the extract_steps step from ~19s to ~10s.

### Bias
My test was using 100 copies of my new steel axe run, which obviously isn't a real example of a run. It is majorly skewed towards take and put tasks.

### DIFF
--- a/src/Functions.h
+++ b/src/Functions.h
@@ -20,8 +20,8 @@ using std::stringstream;
 
 bool check_input(const string& item, const vector<string>& all_items);
 
-string inline Capitalize_std(const std::string& stringToCapitalize, const bool isItem = false);
-string inline Capitalize(const wxString& stringToCapitalize, const bool isItem = false);
+string Capitalize(const std::string& stringToCapitalize, const bool isItem = false);
+string Capitalize(const wxString& stringToCapitalize, const bool isItem = false);
 
 bool equals_ignore_case(const std::string_view& lhs, const std::string_view& rhs);
 bool starts_with_ignore_case(const std::string& base, const std::string& start);
--- a/src/Functions.h
+++ b/src/Functions.h
@@ -20,7 +20,8 @@ using std::stringstream;
 
 bool check_input(const string& item, const vector<string>& all_items);
 
-string Capitalize(const wxString& stringToCapitalize, const bool isItem = false);
+string inline Capitalize_std(const std::string& stringToCapitalize, const bool isItem = false);
+string inline Capitalize(const wxString& stringToCapitalize, const bool isItem = false);
 
 bool equals_ignore_case(const std::string_view& lhs, const std::string_view& rhs);
 bool starts_with_ignore_case(const std::string& base, const std::string& start);
--- a/src/OpenTas.cpp
+++ b/src/OpenTas.cpp
@@ -138,11 +138,11 @@ Category OpenTas::extract_steps(std::ifstream& file, DialogProgressBar* dialog_p
 			segment_size != step_segment_size_without_comment_and_colour &&
 			segment_size != step_segment_size_without_comment_and_colour_and_modifiers)
 		{
-			if (segments[0] == save_templates_indicator)
+			if (segment_size >= 0 && segments[0] == save_templates_indicator)
 			{
 				return Template;
 			}
-			else if (segment_size == 1 && (segments[0] == save_groups_indicator))
+			else if (segment_size >= 0 && segments[0] == save_groups_indicator)
 			{
 				return Group;
 			}
@@ -151,7 +151,7 @@ Category OpenTas::extract_steps(std::ifstream& file, DialogProgressBar* dialog_p
 				return Invalid;
 			}
 		}
-			
+
 		StepParameters step(invalidX, 0);
 
 		if (segments[1] != "")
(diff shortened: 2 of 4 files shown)


src/OpenTas.cpp
```diff
@@ -130,67 +130,47 @@ Category OpenTas::extract_steps(std::ifstream& file, DialogProgressBar* dialog_p
...
+			segment_size != step_segment_size_without_comment_and_colour &&
+			segment_size != step_segment_size_without_comment_and_colour_and_modifiers)
 		{
 			if (segments[0] == save_templates_indicator)
 			{
 				return Template;
 			}
-
-			if (segments.size() == 1 && (segments[0] == save_groups_indicator))
+			else if (segment_size == 1 && (segments[0] == save_groups_indicator))
```
MortenTobiasNielsen: I am fairly certain there is no reason to have these else if a and else, given that there is return in the if statements.

src/OpenTas.cpp
```diff
@@ -130,67 +130,47 @@ Category OpenTas::extract_steps(std::ifstream& file, DialogProgressBar* dialog_p
...
-		if (segments.size() == step_segment_size || segments.size() == step_segment_size_without_colour)
-		{
-			comment = segments[9];
+			else
+			{
+				return Invalid;
+			}
 		}
-
+
```
MortenTobiasNielsen: It is small, but now that I am anyway asking you to make changes - could you remove these empty spaces?

src/OpenTas.cpp
```diff
@@ -669,7 +649,8 @@ bool OpenTas::update_segment(std::ifstream& file)
 	{
 		data_line.str(line);
 
-		segments = {};
+		segments.clear();
+		segments.reserve(16);
```
MortenTobiasNielsen, line 669: You don't need to do this, the memory is already allocated and clear doesn't deallocate memory. I am pretty sure that it will be optimized away by the compiler, so it most likely doesn't make a difference for performance, but still. :)

src/cMain.cpp
```diff
@@ -1578,7 +1578,7 @@ void cMain::PopulateStepGrid()
 	}
 
 	size_t steps = StepGridData.size();
-
+	grid_steps->BeginBatch();
```
MortenTobiasNielsen, line 1669: I wonder what this Begin/EndBatch is actually doing to give almost a 50% improvement.

src/Functions.h
```diff
@@ -20,7 +20,8 @@ using std::stringstream;
 
 bool check_input(const string& item, const vector<string>& all_items);
 
-string Capitalize(const wxString& stringToCapitalize, const bool isItem = false);
+string inline Capitalize_std(const std::string& stringToCapitalize, const bool isItem = false);
```
MortenTobiasNielsen: Good catch. I hadn't thought about that we don't use wxString for most of the heavy processing anymore.

MortenTobiasNielsen (changes requested):
Really nice performance uplifts- good work.

src/OpenTas.cpp
theis999: I believe you are correct, however i also think that it reads cleaner this way.

src/OpenTas.cpp
theis999: You might be correct, but the [documentation ](https://cplusplus.com/reference/vector/vector/clear/)doesn't provide a clear answer:

> image

As i read it, it depends on the implementation of the compiler.

src/cMain.cpp
theis999: It [prevents painting](https://docs.wxwidgets.org/3.0/classwx_grid.html#ae51929a4e2843e246abe01d79c5a9183) of the gui element.

src/Functions.h
theis999: Thanks! It definitely wasn't obvious that it would perform so poorly. Obviously it won't matter in most instances, except this scenario where it is done at least 5 times on every row for thousands of rows.

I could look through the code to see if this faster method, should be used elsewhere.

src/OpenTas.cpp
theis999: I added a guard statement to `return Template;` too, so it is consistent.

src/Functions.h
MortenTobiasNielsen: Come to think about it - shouldn't this be an overload of "Capitalize", so the code uses the one which matches the type?

src/OpenTas.cpp
MortenTobiasNielsen: I see - that is kind of confusing. I would guess for performance purposes that the capacity won't be freed, but yeah it is really not clear from the snippet.

src/Functions.h
theis999: Ah you are correct

MortenTobiasNielsen (approved):

PR #299, opened 2023-06-17
Fixed capitalization of warning for no-order

theis999:
Not exactly sure how these issues slipped through 😪


MortenTobiasNielsen (approved):
Well the IDE is not really too helpful, so nice that you found the bug. :)

PR #301, opened 2023-06-18
Refactor folder design

theis999:
### Preview
image

### Description
Created **Data** folder, for const data.
Created **Shared functions** folder, for utility methods.
Created **Structures** folder, for structs and small classes.
Created **TAS save file** folder, for dealing with the .txt tas file.

### Navigation
This doesn't change solution view much, but does impact the folder view. It does however mean that `#include` need to use folder path navigation. Which does look a bit weird when having to go up a folder, for instance `#include "..\Data\utils.h"`.

### Conclusion
I am not sure i like this, but filters doesn't appear to work consistently. So maybe this is the best solution, and it can allow for splitting longer files into more files. `Utils.h` is the prime file for being split.


MortenTobiasNielsen (approved):
Not really sure I am a fun of this either and at least I think it is a little confusing that the TAS save file folder contains the OpenTas files, but we can always change it a little later - luckily moving files around and changing names isn't difficult, but finding the right structure and names most certainly can be. :)

PR #309, opened 2023-06-28
Fixed that take & put using orientation

theis999:
Fixed that take & put used the orientation field and that double clicking them loaded their inventory into that field.

### DIFF
--- a/src/Structures/ParameterChoices.h
+++ b/src/Structures/ParameterChoices.h
@@ -57,8 +57,8 @@ const struct parameter_choices_struct
 
 	//building interactions
 	const int build = building | item,
-	take = building | container,
-	put = building | container,
+	take = building - building_orientation | container,
+	put = building - building_orientation | container,
 	rotate = building | amount,
 	limit = building | amount,
 	priority = building - building_orientation | priority_io,


MortenTobiasNielsen:
I think I added build orientation for a reason. Are you sure it isn't needed for uneven buildings (e.g. boilers)?

theis999:
Oh that is possible. I will test it.

theis999:
I did a bit of testing and couldn't see anything breaking.

- We match buildings using their position point which is not necessarily the same position the building will be in game (that could be an issue, but unrelated)
- There are no buildings that change position due to rotation. Non-square entities can only rotate in 2 orientations after being placed. 
- Underground belts are special, as they behave the same as non-square entities due to them having a flow direction.
- It doesn't appear possible to place an inverted underground belt before or after this change, and i don't think it should be.

There are a few entity types that changes collision rectangle when they collide with other entities. Which changes the distance calculation. Most noteworthy of these is underground pipes connecting to another fluid box.

Somewhat related, our current setup doesn't work for moving vehicles.

MortenTobiasNielsen (approved):

PR #315, opened 2023-06-28
Refactor StepType

theis999:
Removed StepNameToEnum.cpp
Renamed StepNameToEnum.h to StepType.h
Unrolled multiline item
Added non-capitalized version mapped strings
Changed MapStepNameToStepType map to unordered_map for better performance


MortenTobiasNielsen:
I am not really a fan of adding error types instead of ensuring that it isn't possible to get into that state.

theis999:
It isn't exactly easy to remove all errors :) 

When searching for string that isn't in the map MapStepNameToStepType. If search is done using `MapStepNameToStepType.find(step)` then it will return `map.end()`, however if done using `MapStepNameToStepType[step]` then it will return 0. 
Which i realize is irrelevant since `MapStepNameToStepType[step]` is not possible as the map is declared const.

So i suppose i could remove the alignment / error stuff.

theis999:
I have removed the error types & the alignment stuff.
I have unrolled the multiline items so it looks better.
I have changed map to unordered_map as it should perform better.
I have added non-capitalized versions of the step type strings to the map.

MortenTobiasNielsen (approved):
Nice. I cannot see how we would get a step type not existing, so I will merge it in, but would you ensure that the error is handled somewhere. :)

PR #318, opened 2023-06-30
Added btn to handle force modifier

theis999:
Added button to handle adding force to multiple steps. Basically a copy of no-order.

I won't use to much time on making the code pretty now as i believe most of it is going to change soon.


MortenTobiasNielsen:
Why is this needed?

theis999:
I personally don't like it either. However Zaspar begged for it, so i made it.

MortenTobiasNielsen:
For some reason I don't remember what force is - is it override? Why does Zaspar want a button instead of the check box?

theis999:
Yes, it skips the reach verification which means that there are no intermediate walks steps. It might also skip the verification that a step is connected to a building that exists. 
He prefers just having a start & end point, instead of having to walk steps between them. For building larger structures. So he started putting force on 50+ steps in a row.

MortenTobiasNielsen:
That does seem like the way to do it. The only problem is to time it, so the character don't stand still. That is though another problem. 

So he wants a button - are there any reason to have both a check box and a button?

theis999:
I feel like that is his own problem (at least for now)

Yes, the checkbox holds the state of the current step. Which is useful if the user doesn't click the force button, for instance if they click Add step.

MortenTobiasNielsen (approved):

PR #325, opened 2023-07-13
Template overhaul

theis999:
### Overhaul
This mixes a gui overhaul of the template page, a refactor of code related to templates and adds some new features for templates.

### Changes
Moved most Template stuff from cMain.cpp to a new file called TemplatePage. Refactored most Template functions to follow.

Renamed box sizer in template page-
Changed labels of Template page control buttons-
Added Tooltips to Template page control buttons.
Added Unicode arrows to Import&Export buttons.
Added add step button to template page.

Added Template iterator transformation.
Added Template direction transformation.
Changed amount transformation to allow both types at the same time.

Added button toggles.

image

### DIFF
--- a/src/Factorio-TAS-Generator.vcxproj
+++ b/src/Factorio-TAS-Generator.vcxproj
@@ -180,6 +180,7 @@
     <ClCompile Include="Structures\StepParameters.cpp" />
     <ClCompile Include="TAS save file\OpenTas.cpp" />
     <ClCompile Include="TAS save file\SaveTas.cpp" />
+    <ClCompile Include="TemplatePage.cpp" />
     <ClCompile Include="TypePanel.cpp" />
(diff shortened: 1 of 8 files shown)


MortenTobiasNielsen:
I unfortunately won't have time to test the functionality out for at least a week, but if you have tested it and think it works I will approve it. It first need to really work when it is released. :)

theis999:
I have tested each part in insolation, and they work for the scenarios i can imagine. The scenarios i can't imagine, who knows ;)
But yes i believe it works.

One thing of note. Since the new rotation system assumes Origo as the pivot point, that means it probably can't be combined with the iteration system.

MortenTobiasNielsen:
Seems reasonable that the user has to do those separately. 👍

MortenTobiasNielsen (approved):

PR #336, opened 2023-08-20
Refactor stepparameters

theis999:
## Refactor of StepParameters

Many small changes to make StepParameters better aligned with my personal preference. Which in brief is to use specialized types for fields, which also avoids many situations where the field holds invalid data.

Given the large number of changes it might be easier to look through them per commit.

1. Renamed the field "StepEnum" to "type"
2. Renamed the field "FromInto" to "inventory"
3. Changed the field type `inventory` to `InventoryType`
4. Added aliases fields to GridEntry.h
5. Split `cMain::UpdateParameters()` into two functions
6. Fixed double clicking on step of the type `game_speed`
7. Moved initialization of StepParameter fields to the header file.
8. Changed the type Colour from `string` to `wxColour`, using wxNullColour as unset instead of empty string
9. Renamed the field `Orientation` to `orientation`, as that caused conflicts with the type Orientation
10. Changed the field type of `direction` from `string` to `Orientation`
11. Added new file for shared string functions

### Shared string functions
This change is somewhat half baked, as i didn't move functions from functions.h to it

However it is also necessary, as the inclusion of our own base types such as Orientation, means that functions can't be included to improve Orientation.h as that gives compiler loops.

### Aliases
The uses of reference fields allows us to have alias field. It works a bit like pointers that can't be re-pointed. Where any change to either the referenced or the referencing field is reflected either place.

### Follow up
I still have a few changes before i am satisfied with StepParameters. 

- I want it renamed to Step, but changing that creates so many changes that it is impossible to see what else has changed
- The field Amount should be a number type
- The field Item should be something more appropriate
- The field orientation should be of the Orientation type

### DIFF
--- a/src/Factorio-TAS-Generator.vcxproj
+++ b/src/Factorio-TAS-Generator.vcxproj
@@ -174,6 +174,7 @@
     <ClCompile Include="GUI_Base.cpp" />
     <ClCompile Include="ImportStepsPanel.cpp" />
     <ClCompile Include="Shared functions\Functions.cpp" />
+    <ClCompile Include="Shared functions\StringFunctions.cpp" />
     <ClCompile Include="ShortcutChanger.cpp" />
     <ClCompile Include="SearchUtil.cpp" />
     <ClCompile Include="SteptypeColour.cpp" />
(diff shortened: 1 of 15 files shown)


MortenTobiasNielsen (approved):
seems like some nice additions. Good idea not to rename. :)

PR #337, opened 2023-08-25
Fix that mining takes 1 additional tick when the step is not ended by an event.

theis999:
I am somewhat curious why this line was ever added.

The game will automatically disable mining in the next tick.

_______________________________

So it turned out that the line was there to prevent skipping the next step.

### DIFF
--- a/Lua Files/control.lua
+++ b/Lua Files/control.lua
@@ -1525,7 +1525,7 @@ local function handle_ontick()
 			ticks_mining = ticks_mining + 1
 
 			if ticks_mining >= duration then
-				player.mining_state = {mining = false, position = steps[step][3]}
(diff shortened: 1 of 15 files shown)


theis999:
Not sure about compatibility

theis999:
This is bugged if you finish mining in the same tick as you stop mining.

theis999:
I have updated the solution,
It is a bit more complex now as i have accounted for sustaining the legacy version of mining.
I used log config as inspiration for the solution.

### Changes
Added new top menu : Legacy
Added menu item : Legacy Mining
Added struct : generate_config
Added generate_config to open/save tas
Added LEGACY_MINING to variables.lua


### Result
This change allows toggling legacy mining
Without legacy mining, mining steps are 1 tick faster

MortenTobiasNielsen (approved):

PR #339, opened 2023-08-29
Added scenario Supply challenge.

### LINKED ISSUE #335: Support for Supply challenge scenario
Hello,

I'm interested in creating TAS for Factorio's Supply challenge scenario.

However there's couple of things that would need to be considered and changed in FTG.

- Apart from slight changes in `control.lua` there is no problem in running FTG generated TAS with the Supply challenge scenario.
- `goals.lua`would need to be extended as well, since Supply challenge features custom win condition.
- The main problem is with advancing through the scenario levels: in RTA you do that by clicking a `mod-gui` button, which is not possible through lua API. There is however a way to workaround this. In FTG there would need to be a new step type "Next level" (similiar to "Launch"). This would call a remote interface that would advance the level in the scenario. HOWEVER such a remote doesn't exist, leading to a three available solutions:
    1. File an official feature request to Factorio devs and have them add such a remote interface (not really that hard, but depends on their will).
    2. Using Factorio's [Instrument mode](https://lua-api.factorio.com/latest/Instrument.html) hijack the scenario's `control.lua` and add the remote ourselves. See code below, this would be an extra lua file `instrument-control.lua` that would need to be a part of the generated mod. Then the game needs to be run with the command line argument `--instrument-mod modname`, which is not ideal. It also disables multiplayer, which only affects the Supply challenge scenario in a way that in SP it shows extra messages at the start that need to be manually skipped.
    3. Use custom/edited Supply challenge scenario for the TAS purposes providing said interface.

```lua
mod_gui = require "mod-gui"

if script.mod_name == "level" and script.level and script.level.level_name == "supply" then
    remote.add_interface("DunRaider-TAS-supply",
    { next_level = function(player_index)
            -- fake the "next level" button click event
            local player = game.get_player(player_index)
            local btn = mod_gui.get_button_flow(player).next_level
            if (not btn) or (not btn.enabled) then return end
            local event = {
                name=1,
                tick=game.tick,
                player_index=player_index,
                element=btn,
            }
            story_update(global.story, event, "")
        end,
    })
end
```

Thank you for your consideration and for developing and maintaining this tool.

theis999:
- Added scenario "Supply challenge" supply.
- Added steptype Next to progress in supply.

### Supply
Supply challenge is listed under goals
Choosing supply as goal will enable Next, because i couldn't get WX to play nice with hiding/showing it.
Generating a script with supply as goal, will also move the files needed for supply challenge into the mod.
The files for supply challenge is taken from Factorio's git. With the exception that i didn't include locale for anything but English.


MortenTobiasNielsen:
Is this all set, tested and ready to be merged?

theis999:
Yes. It should be correct (now)

MortenTobiasNielsen (approved):

PR #347, opened 2023-09-23
Refactor open tas

theis999:
### Refactor OpenTas

- Moved enum **Category** into a Struct, to prevent collisions.
- Moved open_file_return_data::auto_close fields into a struct.
- Moved open_file_return_data::auto_put fields into a struct.
- Added method **ReadStep** to consolidate extract step data code from groups, templates, & steps.

__________

Not really happy with the name ReadStep, however i am proud of the technique of using a vector iterator to reuse the same code for steps, templates (and groups).

### DIFF
--- a/src/TAS save file/OpenTas.cpp
+++ b/src/TAS save file/OpenTas.cpp
@@ -25,20 +25,21 @@ open_file_return_data* OpenTas::Open(DialogProgressBar* dialog_progress_bar, std
 		return &return_data;
 	}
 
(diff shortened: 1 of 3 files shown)


MortenTobiasNielsen (approved):
I have a feeling that ReadStep can be simplified and I am not entirely happy about the "buildingsInSnapShot > -1" logic, but given that you are currently the only developer it is more important that you find it intuitive than I do. :D

theis999:
Those are both fair points, and i hope to do a second revision someday.

PR #359, opened 2023-10-02
Changed version

theis999:
- Changed version to 1.0.0
- Enabled multiprocessing of build on release config.

It seems a fine time to call this a proper product.

### DIFF
--- a/src/Data/utils.h
+++ b/src/Data/utils.h
@@ -15,7 +15,7 @@ static const double invalidX = 0.3543534;
 static const struct
 {
 	string name = "Factorio TAS Generator";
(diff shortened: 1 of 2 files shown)


MortenTobiasNielsen:
I don't feel comfortable changing this til a major version, given that I haven't used it for a long time.

theis999:
Would have a minor version be alright? as in 0.4.0

MortenTobiasNielsen (approved):
Yes, a minor is fine - should probably have been a few of those, but given that no release is made I guess it doesn't really change much. 🤷

PR #369, opened 2023-10-22
Complete Undo / redo revamp.

### LINKED ISSUE #350: More undo
There are a few glaring omissions from the undo/redo system:

- Export template, from the template page
- Import steps, from step importer page
- Mark multiple rows for Force or Split

theis999:
### Complete Undo / redo revamp.

- Changed the Command stack size to 128 from 32.
- Changed Commands to have a before,after. Simplifying the undo system. Added new struct StepLine
- Added undo to Templates
- Added undo to walk panel
- Added undo to import steps
- Added undo to force,skip,no_order buttons

Changed MoveRows to use the new undo system as its base function.

Changed BackgroundColorUpdate to also set step colour.

### New system
This new system is way simpler. You take a snap shot of what will be changed and another snap shot of what it was changed into. Which means it can back and forth between the before and after state.

A simple case is delete, where the before part will be the selected steps. While the after part is nothing.

### Templates
Templates are basically the same as the regular steps. However to make it work properly i needed to load the undone template into the template grid. Eventually i realized i would need to create destroyed templates, however undo does not delete templates.
____

This took longer than i care to admit, but the result speaks for itself.

### DIFF
--- a/src/CommandStack.cpp
+++ b/src/CommandStack.cpp
@@ -9,6 +9,7 @@ void CommandStack::Clear()
 
 void CommandStack::Push(Command command)
 {
+	if (command.empty()) return;
 	buffer[head] = command;
 	head = (head + 1) % size;
 	hair = head;
@@ -18,7 +19,7 @@ void CommandStack::Push(Command command)
 Command CommandStack::Pop()
 {
 	if (head == tail) 
-		return {0, T_NULL, {}};
+		return {{}, {}, ""};
 	head = (head + size - 1) % size;
 	return buffer[head];
 }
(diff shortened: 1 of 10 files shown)


theis999:
### Updated
Moved StepLine struct into new file.
Added new StepBlock struct to that file too.
Moved VectorToBlocks to new file too.

Added undo to change steps colour.
Added undo to SplitMultibuildStep.

Fixed type date_list.
Changed undo to also handle focus-mode and warning-states
Chanded undo to use grid batch for better performance.
____

I think this marks undo / redo as feature complete.

